PR #21, opened 2020-03-18
Feature: Add discovery / Reload

Horizon0156:
Hey, hope you're doing well :)

Next one and this time to `dev`. As already mentioned, I've implemented a first version of a reload / discover service. Basically I've connected a service that triggers a reload of all applications and disposes existing daemons. I was thinking about a File watcher first, but as user specific apps might write files to the local folder, this would cause unnecessary reloads. Using this approach the changes are not that complex and the functionality is more convenient than restarting the whole daemon. 

In addition, I've added a new IsEnabled flag that recovers from the corresponding Home Assistant. There it is now possible to enable / disable apps without moving them in the `apps` folder. This is heavily used in my production setup when I temporary need to disable motion sensor triggered automations in case of a full house ;)

The DaemonHost is also writing his state to HA's state machine, so you can easily review the number of loaded / instantiated apps. 

Let me know what you think and feel free to include this to your branch as well. 

I've also published a set of my applications that I've written in the last days in case you want to extend your example list. There are also extensions that deals with Homematic devices and keypress events.

https://github.com/Horizon0156/netdaemon-apps/tree/master/apps 

e.g.
```
HomematicSwitchEvent("000858A99D84DE")
                .Channel1()
                .PressedShort()
                .UseEntities("light.kuche")
                .TurnOn()
                .WithAttribute("brightness", 127)
                .Execute();
```

Thanks again for the great work on this daemon. It's running for a week now in our house without any delays or issues.

Have a good one.  

Preview

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 TestResults
 lcov.info
 codecover
+**.DS_Store
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,5 @@
+ARG RUNTIME_IMAGE_TAG=3.1
+
 FROM mcr.microsoft.com/dotnet/core/sdk:3.1 AS build-env
 WORKDIR /sources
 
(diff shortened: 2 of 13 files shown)


helto4real:
Fantastic! I will check it out ASAP! And thank you for sharing your work with NetDaemon.

helto4real:
Great work with the new features @Horizon0156 !

helto4real:
image
This is my apps running and showing nicely in Home Assistant.. Good job again!

Horizon0156:
Cool :)

Thanks, I'm glad you like it. I will let you know, If I have something new in my mind, but for now the Daemon is running great in my setup.

PR #36, opened 2020-03-23
fix tag action

ludeeus:


ludeeus:
@helto4real This works, no need to wait for builds, just get it in

PR #72, opened 2020-04-12
Builds with correct arch

helto4real:


Dockerfile
```diff
@@ -1,6 +1,11 @@
 # Build the NetDaemon with build container
 FROM mcr.microsoft.com/dotnet/core/sdk:3.1.200-alpine as build
```
ludeeus, line 2: This is "old", there is a 201 now

PR #158, opened 2020-07-08
Pull config out of RunnerService

asherw:
## Breaking change
I believe I've made this compatible with existing config & environment variables.

## Proposed change
Attempting to use built-in .net core host configuration framework.

## Type of change
No new features.

- [ ] Dependency upgrade

## Additional information

## Checklist


helto4real:
Since @BeeHiveJava had some thoughts I wait merging it until he had the chance to look at it too. Thanks!

src/DaemonRunner/DaemonRunner/Service/Extensions/HostBuilderExtensions.cs
```diff
@@ -8,14 +14,33 @@ public static class HostBuilderExtensions
...
                 .UseNetDaemonSerilog()
+                .ConfigureAppConfiguration(configurationBuilder =>
+                {
+                    if (File.Exists(configFile))
+                        configurationBuilder.AddJsonFile(AppSettingsFile);
+                    else
+                        CreateSampleConfigFile();
+
+                    configurationBuilder.AddEnvironmentVariables();
+                })
```
BeeHiveJava, line 33: Is it possible to move this to its own extension method? I think that would stop the temptation of adding more configuration in here the future.

src/DaemonRunner/DaemonRunner/Service/Extensions/HostBuilderExtensions.cs
```diff
@@ -8,14 +14,33 @@ public static class HostBuilderExtensions
...
                 .ConfigureServices(services =>
                 {
                     services.AddNetDaemon();
+                    services.AddSingleton(provider =>
+                    {
+                        var configuration = provider.GetService<IConfiguration>();
+                        var factory = new HostConfigFactory(configuration);
+
+                        return factory.Create();
+                    });
```
BeeHiveJava, line 43: Same thing here, moving this behind the already existing ```AddNetDaemon``` extension method would be clearer I think.

Also I think a more accepted way of injecting configuration is by using ```ConfigureOptions``` and using ```I[Post]Configuration``` to actually load the configuration (the code in ```HostConfigFactory```). This way classes can make use of the ```IOptions``` interface. This also has the added benefit that we can set configuration from code, which is useful during testing. What do you think?

src/DaemonRunner/DaemonRunner/Service/Infrastructure/Configuration/HostConfigFactory.cs
```diff
@@ -0,0 +1,71 @@
...
+        public HostConfigFactory(IConfiguration configuration)
+        {
+            _configuration = configuration;
+        }
+
+        public HostConfig Create()
+        {
+            var config = new HostConfig();
+
+            var host = GetSetting(_configuration, "Host", "HASS_HOST");
```
BeeHiveJava, line 18: Is ```Host``` supposed to start with a capital ```H``` here?

src/DaemonRunner/DaemonRunner/Service/Infrastructure/Configuration/HostConfigFactory.cs
```diff
@@ -0,0 +1,71 @@
...
+            var port = GetSettingShort(_configuration, "port", "HASS_PORT");
+            var token = GetSetting(_configuration, "token", "HASS_TOKEN");
+            var sourceFolder = GetSetting(_configuration, "source_folder", "HASS_DAEMONAPPFOLDER");
+            
+            config.Ssl = GetSettingBool(_configuration, "ssl", "HASS_SSL").GetValueOrDefault(false);
+            config.GenerateEntitiesOnStartup = GetSettingBool(_configuration, "generate_entities", "HASS_GEN_ENTITIES").GetValueOrDefault(false);
+
+            if (string.IsNullOrEmpty(host))
+                return null!;
```
BeeHiveJava, line 27: Can we do this check before setting the ```Ssl``` and the ```GenerateEntitiesOnStartup```?

src/DaemonRunner/DaemonRunner/Service/Infrastructure/Configuration/HostConfigFactory.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            return short.Parse(theValue);
+        }
+
+        private string GetSetting(IConfiguration configuration, string configName, string envName)
+        {
+            if (!string.IsNullOrEmpty(configuration[envName]))
+                return configuration[envName];
+
+            return configuration[configName];
```
BeeHiveJava, line 68: I have a feeling that these dictionaries might throw exceptions when the keys don't exist. Would it be better to either use ```ContainsKey``` or ```TryGet``` methods (in combination with the null check of course)?

src/DaemonRunner/DaemonRunner/Service/Infrastructure/Configuration/HostConfigFactory.cs
```diff
@@ -0,0 +1,71 @@
...
+            if (!string.IsNullOrEmpty(token))
+                config.Token = token;
+            if (!string.IsNullOrEmpty(sourceFolder))
+                config.SourceFolder = sourceFolder;
+
+            return config;
+        }
+
+
+        private bool? GetSettingBool(IConfiguration configuration, string configName, string envName)
```
BeeHiveJava, line 43: I like this approach of the settings combined with environment variables. Can we move this logic to its own class (possibly with interface) so that we can reuse it for other config files (```HassioConfig``` for example)?

src/DaemonRunner/DaemonRunner/Service/RunnerService.cs
```diff
@@ -25,13 +23,15 @@ public class RunnerService : BackgroundService
 
         private const string _version = "dev";
         private readonly IHttpClientFactory _httpClientFactory;
+        private readonly HostConfig _configuration;
 
         private readonly ILogger<RunnerService> _logger;
 
         private readonly ILoggerFactory _loggerFactory;
-        public RunnerService(ILoggerFactory loggerFactory, IHttpClientFactory httpClientFactory)
+        public RunnerService(ILoggerFactory loggerFactory, IHttpClientFactory httpClientFactory, HostConfig configuration)
```
BeeHiveJava, line 31: As said before, I think we can make use of the ```IOptions``` interface here instead of injecting the HostConfig directly.

src/DaemonRunner/DaemonRunner/Service/RunnerService.cs
```diff
@@ -112,7 +102,7 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
                                     try
                                     {
                                         // Generate code if requested
-                                        var envGenEntities = Environment.GetEnvironmentVariable("HASS_GEN_ENTITIES") ?? config.GenerateEntitiesOnStartup?.ToString();
+                                        var envGenEntities = Environment.GetEnvironmentVariable("HASS_GEN_ENTITIES") ?? _configuration.GenerateEntitiesOnStartup?.ToString();
```
BeeHiveJava, line 105: Isn't this check for the environment variable already done implicitly now when loading the configuration?

helto4real:
Give me a ping when it is ready to merge :)

src/DaemonRunner/DaemonRunner/Service/Runner.cs
```diff
@@ -75,5 +46,34 @@ public static async Task Run(string[] args)
...
+                
+                if (hassAddOnSettings.LogLevel is object)
+                    SerilogConfigurator.SetMinimumLogLevel(hassAddOnSettings.LogLevel);
+
+                if (hassAddOnSettings.GenerateEntitiesOnStart is object)
+                    Environment.SetEnvironmentVariable("NetDaemon__GenerateEntities", hassAddOnSettings.GenerateEntitiesOnStart.ToString());
+
+                // I couldn't find any usages.
+                //if (hassAddOnSettings.LogMessages is object && hassAddOnSettings.LogMessages == true)
+                //    Environment.SetEnvironmentVariable("HASSCLIENT_MSGLOGLEVEL", "Default");
```
BeeHiveJava, line 65: Used by a different library (HassClient).

src/DaemonRunner/DaemonRunner/Service/Configuration/HomeAssistantSettings.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace NetDaemon.Service.Configuration
+{
+    public class HomeAssistantSettings
+    {
+        public string Host { get; set; } = "localhost";
+        public short Port { get; set; } = 8123;
+        public bool Ssl { get; set; } = false;
+        public string Token { get; set; } = "";
```
BeeHiveJava, line 8: ```string.Empty``` for clarity.

src/DaemonRunner/DaemonRunner/Service/Runner.cs
```diff
@@ -13,58 +15,27 @@ public static class Runner
     {
         private const string _hassioConfigPath = "/data/options.json";
 
-        public static IHostBuilder CreateHostBuilder(string[] args) => 
-            Host.CreateDefaultBuilder(args)
-                .UseNetDaemon();
-        
         public static async Task Run(string[] args)
```
BeeHiveJava, line 18: I don't like that this prevents us from accessing the host builder. We can't inject our own services anymore now. I think it's better to provide extension methods (in addition to the run method).

src/DaemonRunner/DaemonRunner/Service/Runner.cs
```diff
@@ -13,58 +15,27 @@ public static class Runner
...
-                    {
-                        Log.Fatal(e, "Failed to read the Home Assistant Add-on config");
-                    }
-                }
-                else
-                {
-                    var envLogLevel = Environment.GetEnvironmentVariable("HASS_LOG_LEVEL");
-                    if (!string.IsNullOrEmpty(envLogLevel))
+                await Host.CreateDefaultBuilder(args)
+                    .UseSerilog(Log.Logger)
```
BeeHiveJava, line 28: Nice! I didn't know this was possible.

src/DaemonRunner/DaemonRunner/Service/Runner.cs
```diff
@@ -13,58 +15,27 @@ public static class Runner
...
+                    .ConfigureServices((context, services) =>
                     {
-                        SeriLogConfigurator.SetMinimumLogLevel(envLogLevel);
-                    }
-                }
+                        services.AddOptions<HomeAssistantSettings>().Bind(context.Configuration.GetSection("HomeAssistant"));
```
BeeHiveJava, line 31: I think we can make use of the configure method here, like so:

```C#
services.Configure<HomeAssistantSettings>(context.Configuration.GetSection("HomeAssistant"));
```

This way we don't have to bind early.

src/DaemonRunner/DaemonRunner/Service/Runner.cs
```diff
@@ -13,58 +15,27 @@ public static class Runner
...
+                        services.AddOptions<NetDaemonSettings>().Bind(context.Configuration.GetSection("NetDaemon"));
```
BeeHiveJava, line 32: Same thing here.

src/DaemonRunner/DaemonRunner/Service/Support/SeriLogConfigurator.cs
```diff
@@ -5,18 +5,15 @@
 
 namespace NetDaemon.Service.Support
 {
-    internal static class SeriLogConfigurator
+    public static class SerilogConfigurator
```
BeeHiveJava, line 8: Do we need this to be public?

asherw:
Example of it running locally.
image

src/Service/appsettings.json
```diff
@@ -0,0 +1,12 @@
+﻿{
+    "HomeAssistant": {
+        "Host": "192.168.0.2",
```
BeeHiveJava, line 3: This doesn't seem like a sensible default to me, let's use ```localhost```?

asherw:
I've noticed a couple of flaky tests locally.
I'll have a look into NetDaemon.Daemon.Tests.NetDaemonApp.FaultyAppTests.ARunTimeErrorShouldLogError now but I'm not expecting to be able to find it.

asherw:
How to override appsetting config values using environment variables.

https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-3.1#environment-variables

Eg, "NetDaemon__GenerateEntities" set to "True"

asherw:
Example of log level working.

I set LOGGING__MINIMUMLEVEL to Warning.
image

No info shown.

.devcontainer/devcontainer.json
```diff
@@ -17,12 +17,12 @@
...
-		"HASS_HOST": "${localEnv:HASS_HOST}",
-		"HASS_PORT": "${localEnv:HASS_PORT}",
+		"HOMEASSISTANT__TOKEN": "${localEnv:HOMEASSISTANT__TOKEN}",
+		"HOMEASSISTANT__HOST": "${localEnv:HOMEASSISTANT__HOST}",
+		"HOMEASSISTANT__PORT": "${localEnv:HOMEASSISTANT__PORT}",
 		"HASS_LOGLEVEL": "${localEnv:HASS_LOGLEVEL}",
-		"HASS_GEN_ENTITIES": "${localEnv:HASS_GEN_ENTITIES}",
-		"HASS_DAEMONAPPFOLDER": "${localEnv:HASS_DAEMONAPPFOLDER}"
+		"NETDAEMON__GENERATEENTITIES": "${localEnv:NETDAEMON__GENERATEENTITIES}",
+		"NETDAEMON__SOURCEFOLDER": "${localEnv:NETDAEMON__SOURCEFOLDER}"
```
helto4real, line 25: There is another setting called HASS_RUN_PROJECT_FOLDER that points to a custom folder where the project file is. If that is provided NetDameon add-on and container uses that project file instead of the default. This enables you to add any kind of additional dependencies (nuget packages) etc and use that in the container or add-on. HASS_RUN_PROJECT_FOLDER could be called NETDAEMON__PROJECTFOLDER.

src/Service/Program.cs
```diff
@@ -1,13 +1,67 @@
-﻿using System.Threading.Tasks;
-using NetDaemon.Service;
+﻿using System;
```
helto4real, line 1: We need to remember to update this change in the template now when moved back the initialize logic to Service. 

https://github.com/net-daemon/netdaemon-app-template/blob/master/program.cs#L6

helto4real: Also we need to check with @ludeeus if this will break his devcontainer somehow.

Dockerfile
```diff
@@ -14,11 +14,11 @@ ENV \
     DOTNET_NOLOGO=true \
     DOTNET_CLI_TELEMETRY_OPTOUT=true \
     HASS_RUN_PROJECT_FOLDER=/usr/src/Service \
```
helto4real, line 16: Should be NETDAEMON__PROJECTFOLDER as a suggestion so we keep same naming convention. Thiss affects https://github.com/net-daemon/netdaemon/blob/dev/Docker/rootfs/etc/services.d/NetDaemon/run#L14

src/DaemonRunner/DaemonRunner/DaemonRunner.csproj
```diff
@@ -26,7 +26,6 @@
     <PackageReference Include="Microsoft.Extensions.Hosting" Version="3.1.5" />
     <PackageReference Include="Microsoft.Extensions.Http" Version="3.1.5" />
     <PackageReference Include="YamlDotNet" Version="8.1.2" />
-    <PackageReference Include="Serilog.AspNetCore" Version="3.2.0" />
```
helto4real, line 29: We need to remember to ad to the dev template. See other comment for link

PR #326, opened 2021-02-24
fix 325: Removed possible race condition when updating InternalState dictionary

### LINKED ISSUE #325: Possible Racecondition while updating InternalState
## The problem

 https://github.com/net-daemon/netdaemon/blob/aa47c3f4b27d0b104b28bbe8cf7e7d5366c97cde/src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs#L787

Here the InternalState is cleared and than filled again, which might cause another thread to see a partially filled dictionary and not find the state for an entity it is looking for. I think a better way would be to create a new dictionary for the new state objects and then update the InternalState field to point to the new dictionary when it is ready. I think doing it this way it does not even need to be a ConcurrentDictionary because reading a normal Dictionary is allready threadsafe
Another way would be to do a smarter marge that avoids initially clearing the InternalStates, but just remove the ones that are no longer in the new set and update / add the ones that are

## Environment

- NetDaemon release with the issue:
- Last working NetDaemon release (if known):
- Operating environment (Home assistant Add-on/Docker/Dev setup):

```c#

```

## Traceback/Error logs

```txt

```

FrankBakkerNl:
## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #325 

## Checklist

### DIFF
--- a/src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
+++ b/src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
@@ -784,7 +784,11 @@ namespace NetDaemon.Daemon
             var initialStates = hassStates.Select(n => n.Map())
                 .ToDictionary(n => n.EntityId);
 
-            InternalState.Clear();
+            foreach (var entityId in initialStates.Keys.Where(k => !initialStates.ContainsKey(k)))
+            {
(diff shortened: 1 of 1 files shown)


helto4real (approved):
Good find, Clear is not thread safe.

PR #334, opened 2021-03-16
Bugfix: Pass correct NetDeamonApp into Entities constructors

### LINKED ISSUE #335: Loading entities from nested yaml does not pass deamonApp int constructor
## The problem

when using th efollowing config and property

```yaml
SyncClimateToWindow:
  class: Heating.SyncClimateToWindowApp
    climates:
    - climate.radiator_valve_badkamer
```

```c#

        public IEnumerable<ClimateEntity> Climates { get; set; }
```

The Climates collection gets initialized with one instance of ClimateEntity, but it has its  netDaemonApp field set to null. This causes it to well, not work.

## Environment

- NetDaemon release with the issue:
- Last working NetDaemon release (if known):
- Operating environment (Home assistant Add-on/Docker/Dev setup):

```c#

```

## Traceback/Error logs

```txt

```

FrankBakkerNl:
## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #335

## Checklist

### DIFF
--- a/src/Daemon/NetDaemon.Daemon/Daemon/Config/YamlAppConfig.cs
+++ b/src/Daemon/NetDaemon.Daemon/Daemon/Config/YamlAppConfig.cs
@@ -96,7 +96,7 @@ namespace NetDaemon.Daemon.Config
 
                     var valueType = entry.Value.NodeType;
 
-                    var instance = InstanceProperty(netDaemonApp, prop.PropertyType, entry.Value);
+                    var instance = InstanceProperty(netDaemonApp, netDaemonApp, prop.PropertyType, entry.Value);
 
                     prop.SetValue(netDaemonApp, instance);
                 }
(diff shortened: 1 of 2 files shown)


src/Daemon/NetDaemon.Daemon/Daemon/Config/YamlAppConfig.cs
```diff
@@ -110,13 +110,13 @@ public IEnumerable<INetDaemonAppBase> Instances
         }
 
         [SuppressMessage("", "CA1508")] // Weird bug that this should not warn!
-        private object? InstanceProperty(object? parent, Type instanceType, YamlNode node)
+        private object? InstanceProperty(INetDaemonAppBase deamonApp, object? parent, Type instanceType, YamlNode node)
```
FrankBakkerNl, line 113: Not sure if the parent is actually still needed seems like it is only used for error reporting and sometimes it is passed as null

src/Daemon/NetDaemon.Daemon/Daemon/Config/YamlAppConfig.cs
```diff
@@ -130,7 +130,7 @@ public IEnumerable<INetDaemonAppBase> Instances
 
                     foreach (YamlNode item in ((YamlSequenceNode)node).Children)
                     {
-                        var instance = InstanceProperty(null, listType, item) ??
+                        var instance = InstanceProperty(deamonApp, null, listType, item) ??
                                     throw new NotSupportedException($"The class {parent?.GetType().Name} has wrong type in items");
```
FrankBakkerNl, line 134: I actually think these error messages are phrased the wrong way around. They tell why the code does not match the yaml. It would make more sense to assume the code is correct and help the author of the configuration how to correct the yaml by eg referencing the line number of the config file instead of the name of the class

helto4real: Good point! If you do not mind you could change that in this PR.

src/Daemon/NetDaemon.Daemon/Daemon/Config/YamlAppConfig.cs
helto4real: Yea that is what its for. To make the error reporting better. Have no good ideas how to refactor this now. It is used when handling scalar objects

helto4real (approved):
It looks good. I will merge this to dev now. If you want to submit a change of the error message it would be appreciated.

PR #337, opened 2021-03-23
Entity Extensions

ngelotte:
## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/src/App/NetDaemon.App/Common/Reactive/Services/BinarySensorEntity.cs
+++ b/src/App/NetDaemon.App/Common/Reactive/Services/BinarySensorEntity.cs
@@ -1,7 +1,52 @@
 ﻿using System.Collections.Generic;
+using System;
+using System.Reactive;
+using System.Reactive.Linq;
 
 namespace NetDaemon.Common.Reactive.Services
 {
+
+    public record BinaryEntityState : EntityState
+    {
+        public BinaryEntityState(EntityState currentEntity) : base(currentEntity)
+        {
+
+        }
+        /// <summary>
+        /// Read the elevation from the attributes of the sun entity. Returns -100 if anything is null or missing
+        /// </summary>
+        /// <value></value>
+        public bool IsOn
+        {
+            get
+            {
+                return State == "on";
+            }
+        }
+
+        /// <summary>
+        /// Returns true if the state equals off
+        /// </summary>
+        public bool IsOff
+        {
+            get
+            {
+                return State == "off";
+            }
+        }
+
+        /// <summary>
+        /// Returns true if the state is null. This means that home assisant cannot contact the devices
+        /// </summary>
+        public bool IsUnknown
+        {
+            get
+            {
+                return State == null;
+            }
+        }
+    }
+
     /// <inheritdoc />
     public class BinarySensorEntity : RxEntityBase
     {
@@ -10,26 +55,43 @@ namespace NetDaemon.Common.Reactive.Services
         {
         }
 
-        public bool IsOn
+        /// <summary>
+        ///  Returns the Current Entity State of the object
+        /// </summary>
+        /// <inheritdoc/>
+        public new IObservable<(BinaryEntityState Old, BinaryEntityState New)> StateAllChanges
         {
             get
             {
-                return (State ?? "Unknown") == "on";
+                return base.StateAllChanges.Select(t => (new BinaryEntityState(t.Old), new BinaryEntityState(t.New)));
             }
         }
-        public bool IsOff
+
+        /// <inheritdoc/>
+        public new IObservable<(BinaryEntityState Old, BinaryEntityState New)> StateChanges
         {
             get
             {
-                return (State ?? "Unknown") == "off";
+                return base.StateChanges.Select(t => (new BinaryEntityState(t.Old), new BinaryEntityState(t.New)));
             }
         }
-        public bool IsUnknown
+
+        public override SwitchEntityState? EntityState
         {
             get
             {
-                return State  == null;
+                if (base.EntityState == null)
+                {
+                    return null;
+                }
+                return (SwitchEntityState)base.EntityState;
             }
         }
+
+
+        public bool IsOn => EntityState.IsOn;
+        public bool IsOff => EntityState.IsOff;
+        public bool IsUnknown => EntityState.IsUnknown;
     }
 }
+ 
--- a/src/App/NetDaemon.App/Common/Reactive/Services/BinarySensorEntity.cs
+++ b/src/App/NetDaemon.App/Common/Reactive/Services/BinarySensorEntity.cs
@@ -3,11 +3,33 @@
 namespace NetDaemon.Common.Reactive.Services
 {
     /// <inheritdoc />
-    public partial class BinarySensorEntity : RxEntityBase
+    public class BinarySensorEntity : RxEntityBase
     {
         /// <inheritdoc />
         public BinarySensorEntity(INetDaemonRxApp daemon, IEnumerable<string> entityIds) : base(daemon, entityIds)
         {
+        }
+
+        public bool IsOn
+        {
+            get
+            {
+                return (State ?? "Unknown") == "on";
+            }
+        }
+        public bool IsOff
+        {
+            get
+            {
+                return (State ?? "Unknown") == "off";
+            }
+        }
+        public bool IsUnknown
+        {
+            get
+            {
+                return State  == null;
+            }
         }
     }
 }
--- a/src/App/NetDaemon.App/Common/Reactive/Services/EntityGroupExtensions.cs
+++ b/src/App/NetDaemon.App/Common/Reactive/Services/EntityGroupExtensions.cs
@@ -54,42 +54,29 @@ namespace NetDaemon.Common.Reactive.Services
         /// Check if all entities are on
         /// </summary>
         /// <param name="entities">List of entities</param>
-        public static bool AreAllOn(this List<SwitchEntity> entities)
-        {
-            if (entities != null)
-            {
-                foreach (SwitchEntity entity in entities)
-                {
-                    if (entity.IsOff)
-                    {
-                        return false;
-                    }
-
-                }
-            }
-            return true;
-        }
+        public static bool AreAllOn(this List<SwitchEntity> entities) => entities.All(e => e.IsOn);
 
 
         /// <summary>
         /// Check if all entities are off
         /// </summary>
         /// <param name="entities">List of entities</param>
-        public static bool AreAllOff(this List<SwitchEntity> entities)
-        {
-            if (entities != null)
-            {
-                foreach (SwitchEntity entity in entities)
-                {
-                    if (entity.IsOn)
-                    {
-                        return false;
-                    }
+        public static bool AreAllOff(this List<SwitchEntity> entities)  => entities.All(e => e.IsOn);
 
-                }
-            }
-            return true;
-        }
+
+        /// <summary>
+        /// Check if all entities are on
+        /// </summary>
+        /// <param name="entities">List of entities</param>
+        public static bool AreAllOn(this List<BinarySensorEntity> entities) => entities.All(e => e.IsOn);
+
+
+        /// <summary>
+        /// Check if all entities are off
+        /// </summary>
+        /// <param name="entities">List of entities</param>
+        public static bool AreAllOff(this List<BinarySensorEntity> entities)  => entities.All(e => e.IsOn);
+
         /// <summary>
         /// Subscribe to all changes in list
         /// </summary>
--- a/src/App/NetDaemon.App/Common/Reactive/Services/EntityGroupExtensions.cs
+++ b/src/App/NetDaemon.App/Common/Reactive/Services/EntityGroupExtensions.cs
@@ -3,6 +3,9 @@ using System.Collections.Generic;
 using System;
 using System.Runtime.CompilerServices;
 using System.Linq;
+using System.Reactive;
+using System.Reactive.Linq;
+
 
 namespace NetDaemon.Common.Reactive.Services
 {
@@ -15,7 +18,7 @@ namespace NetDaemon.Common.Reactive.Services
         /// Turn all entities in the list off. Checks if the entity is on before turning it off.
         /// </summary>
         /// <param name="entities">List of entities to Turn Off</param>
-        public static void TurnAllOff(this IEnumerable<SwitchEntity> entities)
+        public static void TurnOff(this IEnumerable<SwitchEntity> entities)
         {
             if (entities != null)
             {
@@ -34,7 +37,7 @@ namespace NetDaemon.Common.Reactive.Services
         /// Turn all entities in the list on. Checks if the entity is off before turning it on.
         /// </summary>
         /// <param name="entities">List of entities to Turn Off</param>
-        public static void TurnAllOn(this IEnumerable<SwitchEntity> entities)
+        public static void TurnOn(this IEnumerable<SwitchEntity> entities)
         {
             if (entities != null)
             {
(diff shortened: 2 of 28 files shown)


src/App/NetDaemon.App/Common/Reactive/Services/BinarySensorEntity.cs
```diff
@@ -3,11 +3,33 @@
...
     /// <inheritdoc />
-    public partial class BinarySensorEntity : RxEntityBase
+    public class BinarySensorEntity : RxEntityBase
     {
         /// <inheritdoc />
         public BinarySensorEntity(INetDaemonRxApp daemon, IEnumerable<string> entityIds) : base(daemon, entityIds)
         {
         }
+
+        public bool IsOn
```
FrankBakkerNl, line 13: I would do these like
public bool IsOn => State == "on"

the ?? "unknown" is not needed here I think because null does not equal "on"

src/App/NetDaemon.App/Common/Reactive/Services/EntityGroupExtensions.cs
```diff
@@ -0,0 +1,135 @@
...
+                }
+            }
+            return true;
+        }
+        /// <summary>
+        /// Subscribe to all changes in list
+        /// </summary>
+        /// <param name="entities">all entities</param>
+        /// <param name="app">NetDaemonApp</param>
+        public static IObservable<(EntityState Old, EntityState New)> StateAllChanges(this IEnumerable<RxEntityBase> entities, NetDaemonRxApp app)
```
FrankBakkerNl, line 98: I wanted to do this without passing the NetDaemonRxApp app parameter here. I think it should be possible because these entities allready have a refrence to the app, it is just not exposed.

Maybe It can also be done by calling StateChanges on each entity and then combining the IObservables (not sure how that works in RX but I think it would be possible

src/App/NetDaemon.App/Common/Reactive/Services/EntityGroupExtensions.cs
```diff
@@ -0,0 +1,135 @@
...
+                    }
+
+                }
+            }
+
+        }
+
+        /// <summary>
+        /// Check if all entities are on
+        /// </summary>
```
FrankBakkerNl, line 55: This can be written as

public static bool AreAllOn(this IEnumerable<SwitchEntity> entities) => entities.All(e => e.IsOff)

Which also makes this method a bit less usefull I guess

src/App/NetDaemon.App/Common/Reactive/Services/EntityGroupExtensions.cs
```diff
@@ -0,0 +1,135 @@
...
+            var ids = entities.Select(e => e.EntityId).ToArray();
+            return app.Entities(ids).StateChanges;
+        }
+        //EntityState extensions
+        /// <summary>
+        /// Check if an entities state equals on
+        /// </summary>
+        /// <param name="entityState">Current Entity State</param>
+        /// <returns></returns>
+        public static bool IsOn(this EntityState entityState)
```
FrankBakkerNl, line 119: Having IsOn both on the State and the Entity object seems a bit redundant

src/App/NetDaemon.App/Common/Reactive/Services/SunEntity.cs
```diff
@@ -3,11 +3,27 @@
...
 {
     /// <inheritdoc />
-    public partial class SunEntity : RxEntityBase
+    public class SunEntity : RxEntityBase
     {
         /// <inheritdoc />
         public SunEntity(INetDaemonRxApp daemon, IEnumerable<string> entityIds) : base(daemon, entityIds)
         {
         }
+
```
FrankBakkerNl, line 12: You could use an overload for the constructor instead of a default parameter value, like

public SunEntity(INetDaemonRxApp daemon) : this(daemon, new [] { "sun.sun"} )
        {
        }

src/App/NetDaemon.App/Common/Reactive/Services/SwitchEntity.cs
```diff
@@ -3,11 +3,44 @@
...
     {
         /// <inheritdoc />
         public SwitchEntity(INetDaemonRxApp daemon, IEnumerable<string> entityIds) : base(daemon, entityIds)
         {
         }
+
+        /// <summary>
+        /// Returns true if the state equals on
+        /// </summary>
+        public bool IsOn
```
FrankBakkerNl, line 16: Adding these on/offf properties on the specific entities will make it harder to use them generically. 
I am not sure about the HA structure, if 'on / off' is appliccable for all or most entities I would put them on the base class or maybe as extension methods for the base class. Or maybe just leave them as extention method just on the state object.

src/App/NetDaemon.App/Common/Reactive/Services/EntityGroupExtensions.cs
ngelotte: I agree. It would be great to not have to pass the app. But I am not sure the best way to do that. Probably wait to see if anyone else has some good ideas here.

src/App/NetDaemon.App/Common/Reactive/Services/SunEntity.cs
ngelotte: Perfect. I knew I was missing something.

src/App/NetDaemon.App/Common/Reactive/Services/EntityGroupExtensions.cs
ngelotte: The reason I put it here is because then you can using it in the where statements. Like this:
`PirSensors.StateChanges(this)
                        .Where(e => e.New.IsOn())
                        .Subscribe(e => ....`

They are kind of redundent but they serve similar but different purposes in my mind.

src/App/NetDaemon.App/Common/Reactive/Services/BinarySensorEntity.cs
ngelotte: Good thinking. I will update

src/App/NetDaemon.App/Common/Reactive/Services/SwitchEntity.cs
ngelotte: I don't use a lot of the entity types so I don't know how many of them have a concept of on/off. I like that they can be customized to make sense for what the the entity types are supposed to do. They are already kind of the generizied version of that class of thing for Home Assistant. The generated entities would have every property that home assistant exposes. For example the ClimateEntity has it's own concepts, as does Lock and Vacuum. I don't want IsOn or IsOff showing up on things where it would not make sense.

src/App/NetDaemon.App/Common/Reactive/Services/EntityGroupExtensions.cs
ngelotte: Did not think of using All. I like it.

src/App/NetDaemon.App/Common/Reactive/Services/SwitchEntity.cs
FrankBakkerNl: RxEntity has TurnOff() and TurnOff() on it, probably because many entities provide a turn_on / turn_off service. I see in my HAS deveoper tools however this is not the case for all of them.

src/App/NetDaemon.App/Common/Reactive/Services/EntityGroupExtensions.cs
BeeHiveJava: Hey what about `entities.Select(e => e.StateAllChanges).Merge();`.

I don't have time to test this, but it should work. We might want to limit concurrency in the `Merge` operation though.

FrankBakkerNl: I think merge is exacly what I ment. I do not think it will cause any actual concurrency issues. The observabkes that are returned from the StateChange properties of each entity are really just all different filters on the original same observable. Merge will just combine the results of these filters again onto a single stream

FrankBakkerNl: Maybe this method should be called
MergeStateAllChanges()

I found it a bit strange to have a Method in IEnumerable<rxEntityBase> with the same name as a property on some other classes. Starting with a verb also makes more sense for a method.

src/App/NetDaemon.App/Common/Reactive/Services/EntityGroupExtensions.cs
```diff
@@ -0,0 +1,122 @@
...
+    /// <summary>
+    /// Extensions to provide additional features to groups of typed enitites.
+    /// </summary>
+    public static class EntityGroupExtensions
+    {
+        /// <summary>
+        /// Turn all entities in the list off. Checks if the entity is on before turning it off.
+        /// </summary>
+        /// <param name="entities">List of entities to Turn Off</param>
+        public static void TurnAllOff(this IEnumerable<SwitchEntity> entities)
```
helto4real, line 18: I would prefere to have just `TurnOn` cause of context you would know it has to do with many switches given you call your variable like `living_room_lights.TurnOn()` looks nicer imo. I will not comment rest.

src/App/NetDaemon.App/Common/Reactive/Services/EntityGroupExtensions.cs
```diff
@@ -0,0 +1,122 @@
...
+                }
+            }
+
+        }
+
+        /// <summary>
+        /// Check if all entities are on
+        /// </summary>
+        /// <param name="entities">List of entities</param>
+        public static bool AreAllOn(this List<SwitchEntity> entities) => entities.All(e => e.IsOn);
```
helto4real, line 57: Same comment as before. Remove "All"? We keep "Are" so like `if (livingroom_ligts.AreOn(){}`

src/App/NetDaemon.App/Common/Reactive/Services/EntityGroupExtensions.cs
helto4real: I will wait with merging this PR until you confirm it tests ok. But l do not have any idea how "not pass app" ether.

src/App/NetDaemon.App/Common/Reactive/Services/EntityGroupExtensions.cs
helto4real: I agree with Frank on this one. How do they serve different purposes? Think API should kept as small as possible. This is also on the general "EntityState" class. That means that entities that also not implement on/off will have this method.

src/App/NetDaemon.App/Common/Reactive/Services/LightEntity.cs
```diff
@@ -3,11 +3,44 @@
...
     {
         /// <inheritdoc />
         public LightEntity(INetDaemonRxApp daemon, IEnumerable<string> entityIds) : base(daemon, entityIds)
         {
         }
+
+        /// <summary>
+        /// Returns true if the state equals on
+        /// </summary>
+        public bool IsOn
```
helto4real, line 16: Here it makes sense to have "IsOn" since it is a light after all

src/App/NetDaemon.App/Common/Reactive/Services/SwitchEntity.cs
helto4real: Should be ok for individial entities that actually have On/Off states. I rather have  them on those than the State object.

helto4real (changes requested):
Thanks, looks great overall.

I do have some changes, the tests of "merge" part. Just comment or even better write tests for it. Also the "IsOn" on the state class I think is not great since it has all kind of states for entities that does not have on/off states as well. I am prepared to change my mind about that if there are a good reason to.

src/App/NetDaemon.App/Common/Reactive/Services/EntityGroupExtensions.cs
FrankBakkerNl: Part of my point was that because you can also write

if (livingroom_ligts.All(l => l.IsOn){}

I do not think this method is very usefull and maybe better to leave it out.

It is also more clear to me that it means all lights are on reather than maybe at least one, innwhich vase I would write

Lighys.Any(l => l.IsOn)

src/App/NetDaemon.App/Common/Reactive/Services/EntityGroupExtensions.cs
FrankBakkerNl: The Merge() suggested by @BeeHiveJava eliminates the need for the app parameter.

I did take a look at how Merge is implemented. It is a bit more complicated than just filtering on any of the id's because it needs to handle concurrent evens from multiple sources which will not happen in this case. So maybe there is some overhead. Removing the parameter from the api however seems to out weigh the possible overhead to me

BeeHiveJava: I agree with @FrankBakkerNl, concurrency is not an issue here and the overhead outweighs the api design. But we could always pass a parameter to `Merge` to make this choice explicit.

```C#
const int MaxConcurrency = 1; // Move this elsewhere, possibly constants module.
entities.Select(e => e.StateAllChanges).Merge(MaxConcurrency);
```

FrankBakkerNl: Did not know about the max concurrency parameter. I would not bother moving it if 1 is e sensibke value here.

Alternative would be to have the Entity class make its App field accessible via an internal property. Then we can check if they are all the same (which would most likely be the case) and use that. Otherwise we so the merge

ngelotte: I will try it with just Merge. I think the overhead will be minimal. I think I will leave it as the default(no max concurrency parameter) just to be safe and le the RX library function as intended.

FrankBakkerNl: _If you pass an Enumerable or Observable of Observables, you have the option of also passing in an integer indicating the maximum number of those Observables it should attempt to be subscribed to simultaneously. Once it reaches this maximum subscription count, it will refrain from subscribing to any other Observables emitted by the source Observable until such time as one of the already-subscribed-to Observables issues an onCompleted notification._

FrankBakkerNl: Seems this maxConxurrent is indeed not what we need here

src/App/NetDaemon.App/Common/Reactive/Services/BinarySensorEntity.cs
```diff
@@ -1,13 +1,97 @@
...
+        /// <inheritdoc/>
+        public new IObservable<(BinaryEntityState Old, BinaryEntityState New)> StateChanges
+        {
+            get
+            {
+                return base.StateChanges.Select(t => (new BinaryEntityState(t.Old), new BinaryEntityState(t.New)));
+            }
+        }
+
+        public override SwitchEntityState? EntityState
```
FrankBakkerNl, line 79: Should not this be BinaryEntityState?

src/App/NetDaemon.App/Common/Reactive/Services/BinarySensorEntity.cs
```diff
@@ -1,13 +1,97 @@
...
+        {
```
FrankBakkerNl, line 80: Allmost all of these get properties can be written using the => syntax which will make most of them simple one-liners. Try to avoid null checks on non-nullable fields (not sure if this can be null here)

public override SwitchEntityState? EntityState => base.EntityState == null ? null : BinaryEntityState(base.EntityState )

src/App/NetDaemon.App/Common/Reactive/Services/EntityGroupExtensions.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// Turn all entities in the list off. Checks if the entity is on before turning it off.
+        /// </summary>
+        /// <param name="entities">List of entities to Turn Off</param>
+        public static void TurnOff(this IEnumerable<SwitchEntity> entities)
+        {
+            if (entities != null)
+            {
+                foreach (SwitchEntity entity in entities)
+                {
+                    if (entity.IsOn)
```
FrankBakkerNl, line 27: If this check is needed it should be inside entity.TurnOff();

src/App/NetDaemon.App/Common/Reactive/Services/EntityGroupExtensions.cs
```diff
@@ -0,0 +1,145 @@
...
+                }
+            }
+
+        }
+
+        /// <summary>
+        /// Check if all entities are on
+        /// </summary>
+        /// <param name="entities">List of entities</param>
+        public static bool AreOn(this List<SwitchEntity> entities) => entities.All(e => e.IsOn);
```
FrankBakkerNl, line 60: I had another comment on these AreOn / AreOff methods which seems to got lost due to the last commit

I think these methods have very limited value , now you an type
myLights.AreOn
instead of
myLights.All(l = l.IsOn)

I have no problem typing the last, so personally I would not add these small helpers to NetDeamon where they need to be mainataned, documented etc.

src/App/NetDaemon.App/Common/Reactive/Services/EntityGroupExtensions.cs
```diff
@@ -0,0 +1,145 @@
...
+
+
+        /// <summary>
+        /// Check if all entities are off
+        /// </summary>
+        /// <param name="entities">List of entities</param>
+        public static bool AreOff(this List<SwitchEntity> entities) => entities.All(e => e.IsOn);
```
FrankBakkerNl, line 67: Should be IsOff I guess

FrankBakkerNl:
I think most of what was in this PR is now also present in the new HassModel. and #498  #499

PR #339, opened 2021-03-28
Refactor NetDaemonHost.HandleNewEvent

FrankBakkerNl:
## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
+++ b/src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
@@ -77,6 +77,7 @@ namespace NetDaemon.Daemon
         ///     Used for testing
         /// </summary>
         internal ConcurrentDictionary<string, INetDaemonAppBase> InternalAllAppInstances { get; } = new();
+
         private bool _isDisposed;
 
         // Following token source and token are set at RUN
@@ -128,6 +129,12 @@ namespace NetDaemon.Daemon
         public ILogger Logger { get; }
 
         public IEnumerable<INetDaemonAppBase> AllAppInstances => InternalAllAppInstances.Values;
+
+        private IEnumerable<NetDaemonRxApp> NetDaemonRxApps =>
+            InternalAllAppInstances.OfType<NetDaemonRxApp>();
+
+        private IEnumerable<IObserver<RxEvent>>? EventChangeObservers =>
+            NetDaemonRxApps.SelectMany(app => ((EventObservable)app.EventChangesObservable).Observers);
 
         [SuppressMessage("", "CA1721")]
         public IEnumerable<EntityState> State => InternalState.Select(n => n.Value);
@@ -782,7 +789,7 @@ namespace NetDaemon.Daemon
             }
             var hassStates = await _hassClient.GetAllStates(_cancelToken).ConfigureAwait(false);
 
-            foreach (var state in hassStates.Select(s=>s.Map()))
+            foreach (var state in hassStates.Select(s => s.Map()))
             {
                 InternalState[state.EntityId] = state with
                 {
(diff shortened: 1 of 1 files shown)


src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
```diff
@@ -821,193 +828,177 @@ protected virtual async Task HandleNewEvent(HassEvent hassEvent, CancellationTok
...
-                                        observer.OnNext((oldState, newState));
-                                    }
-                                    catch (Exception e)
-                                    {
-                                        observer.OnError(e);
-                                        netDaemonRxApp.LogError(e, $"Fail to OnNext on state change observer. {newState.EntityId}:{newState?.State}({oldState?.State})");
-                                    }
-                                }, token));
-                            }
+                            observer.OnError(e);
```
FrankBakkerNl, line 909: I am not sure if this is actually the corrcet way tp handle errors in Observables. I think OnError should be calles when the producers causes an error, not the observer. The same patteren is repeated here a coupe of times slightly different.

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
```diff
@@ -821,193 +828,177 @@ protected virtual async Task HandleNewEvent(HassEvent hassEvent, CancellationTok
...
+            var oldState = stateData!.OldState!.Map();
+            // TODO: refactor map to take area as input to avoid the copy
+            newState = newState with { Area = GetAreaForEntityId(newState.EntityId) };
+            InternalState[stateData.EntityId] = newState;
 
-                    if (!FixStateTypes(stateData))
+            foreach (var netDaemonRxApp in NetDaemonRxApps)
+            {
+                // Call the observable with no blocking
+                foreach (var observer in ((StateChangeObservable)netDaemonRxApp.StateChangesObservable).Observers)
```
FrankBakkerNl, line 899: I wanted to Create a property StateChangesObservables like I did for the other one, but here the netDaemonRxApp is used for logging. The other events do not log using the app so maybe that is not actually needed

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
```diff
@@ -821,193 +828,177 @@ protected virtual async Task HandleNewEvent(HassEvent hassEvent, CancellationTok
...
-                                        }
-                                        catch (Exception ex)
-                                        {
-                                            Logger.LogError(ex, "Fail to OnError on event observer (service_call)");
-                                        }
-                                        Logger.LogError(e, "Fail to OnNext on event observer (service_call)");
-                                    }
-                                }, token));
-                            }
+                            observer.OnError(e);
```
FrankBakkerNl, line 953: This one is even more complicated, in the catch call OnError in another try catch. If this is really needed (what I doubt) this behaviour can probably extracted in a wrapper for the observer. Otherwise It would be enough to just catch and maybe log all exceptions from eventhanders

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
helto4real: Ok well I need to revisit this again. Not sure. Was a quite a time ago I implemented

helto4real:
@FrankBakkerNl thanks for this. Sorry for being late merging. Been busy times at work. I will check it tomorrow, there are some removed code I need to think about if it is ok. Looks good overall !

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
helto4real, line 909: Now I think I know why I did this design. This is so I will pass any exception back to app so apps can also handle exceptions or they would go silent if I remember correctly

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
helto4real: Yea if we gonna refactor the error handling we need to test it so we make sure all exceptions are logged and the process continue handle messages. But you are probably right.

helto4real (approved):

PR #359, opened 2021-04-25
Test CodeGen of all services

FrankBakkerNl:
## Proposed change

Refactored CodeGen (extracted into smaller methods)
generate service wrappers for all domain (not just those with entities)
generate services wihout an entity_id property as method on the domain class
generate services with an entity_idd property as a method on the entity class
Generate records for services with properties for each field
Generate a single parameter method for services with just one field (besides the Entity_id)

## Type of change

- [ ] Dependency upgrade
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (which adds functionality to an existing integration)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [ ] Code quality improvements to existing code or addition of tests

## Additional information

- This PR fixes or closes issue: fixes #
- This PR is related to issue:
- Link to documentation pull request:

## Checklist

- [ ] The code change is tested and works locally.
- [ ] Local tests pass. **Your PR cannot be merged unless tests pass**
- [ ] There is no commented out code in this PR.
- [ ] I have followed the [development checklist][dev-checklist]
- [ ] The code compiles without warnings (code quality chek)
- [ ] Tests have been added to verify that the new code works.

If user exposed functionality or configuration variables are added/changed:

- [ ] Documentation added/updated for [www.home-assistant.io][docs-repository]

[docs-repository]: https://github.com/net-daemon/docs

PR #360, opened 2021-04-29
make code generation logic injectable

Zorgino:
## Proposed change

Make code generation injectable, so devs can test and share new code generation approaches without 
violating open-closed principle.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This blocks PR : https://github.com/net-daemon/netdaemon/pull/359

## Checklist

### DIFF
--- a/src/DaemonRunner/DaemonRunner/Service/App/ICodeGenerationHandler.cs
+++ b/src/DaemonRunner/DaemonRunner/Service/App/ICodeGenerationHandler.cs
@@ -1,0 +1,10 @@
+﻿using System.Threading.Tasks;
+using NetDaemon.Daemon;
+
+namespace NetDaemon.Service
+{
+    public interface ICodeGenerationHandler
+    {
+        Task GenerateEntitiesAsync(NetDaemonHost daemonHost, string sourceFolder);
+    }
+}
--- a/src/DaemonRunner/DaemonRunner/Service/App/ICodeGenerator.cs
+++ b/src/DaemonRunner/DaemonRunner/Service/App/ICodeGenerator.cs
@@ -1,0 +1,10 @@
+﻿using System.Collections.Generic;
+using JoySoftware.HomeAssistant.Client;
+
+namespace NetDaemon.Service.App
+{
+    public interface ICodeGenerator
+    {
+        string? GenerateCodeRx(string nameSpace, IEnumerable<string> entities, IEnumerable<HassServiceDomain> services);
+    }
+}
--- a/src/DaemonRunner/DaemonRunner/Service/RunnerService.cs
+++ b/src/DaemonRunner/DaemonRunner/Service/RunnerService.cs
@@ -33,6 +33,7 @@ namespace NetDaemon.Service
         private readonly IServiceProvider _serviceProvider;
         private readonly IYamlConfig _yamlConfig;
         private readonly IDaemonAppCompiler _daemonAppCompiler;
+        private readonly ICodeGenerationHandler _codeGenerationHandler;
 
         private bool _entitiesGenerated;
         private IEnumerable<Type>? _loadedDaemonApps;
@@ -47,8 +48,8 @@ namespace NetDaemon.Service
             IOptions<HomeAssistantSettings> homeAssistantSettings,
             IServiceProvider serviceProvider,
             IYamlConfig yamlConfig,
-            IDaemonAppCompiler daemonAppCompiler
-            )
+            IDaemonAppCompiler daemonAppCompiler,
+            ICodeGenerationHandler codeGenerationHandler)
         {
             _ = homeAssistantSettings ??
                throw new NetDaemonArgumentNullException(nameof(homeAssistantSettings));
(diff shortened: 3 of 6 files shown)


src/DaemonRunner/DaemonRunner/Service/App/ICodeGenerationHandler.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Threading.Tasks;
+using NetDaemon.Daemon;
+
+namespace NetDaemon.Service
+{
+    public interface ICodeGenerationHandler
+    {
+        Task GenerateEntitiesAsync(NetDaemonHost daemonHost, string sourceFolder);
```
BeeHiveJava, line 8: Let's add a cancellation token (with default value) here?

FrankBakkerNl: I found the CodeGen to be pretty fast (It is just in an memory operation) So I would not expect it to be needed to canceled during the process

FrankBakkerNl: Or is interface also for the entore process, inclusing the writing of the file(s), In that case it could take a bit longer, but still

src/DaemonRunner/DaemonRunner/Service/App/ICodeGenerator.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Collections.Generic;
+using JoySoftware.HomeAssistant.Client;
+
+namespace NetDaemon.Service.App
+{
+    public interface ICodeGenerator
+    {
+        string? GenerateCodeRx(string nameSpace, IEnumerable<string> entities, IEnumerable<HassServiceDomain> services);
```
FrankBakkerNl, line 8: In my PR I changed these IEnumerables into ILists, Manly to avoid warning about 'multiple enumerations of IEnumerable

src/DaemonRunner/DaemonRunner/Service/App/ICodeGenerationHandler.cs
Zorgino: @BeeHiveJava , cancellation token is a great idea which needs to be added to all async methods.
unfortunately, modifying contracts CT is out of scope as code was copy pasted "as-is" from the original place.

src/DaemonRunner/DaemonRunner/Service/RunnerService.cs
```diff
@@ -47,8 +48,8 @@ public class RunnerService : BackgroundService
             IOptions<HomeAssistantSettings> homeAssistantSettings,
             IServiceProvider serviceProvider,
             IYamlConfig yamlConfig,
-            IDaemonAppCompiler daemonAppCompiler
-            )
+            IDaemonAppCompiler daemonAppCompiler,
+            ICodeGenerationHandler codeGenerationHandler)
```
FrankBakkerNl, line 52: I think strictly adding this constrictor parameter to a public class is a breaking change. Probably not a lot of user-code actually uses this constructor

Zorgino: `RunnerService `derives from `BackgroundService` therefore must not be instantiated explicitly

src/DaemonRunner/DaemonRunner/Service/App/ICodeGenerator.cs
Zorgino: you are totally right. could you please apply those changes after this gets merged?

BeeHiveJava: I'd like to avoid IList and instead use ICollection or preferably even IReadOnlyCollection.

This allows to return collections such as a set as well.

FrankBakkerNl: Yeah, your right, I see now I actually did use IReadOnlyCollection here in my PR :-) and did a ToList() on the call-site. This is what caused my PR to initially not pass the CI build

helto4real: That is a good idea using IReadOnlyCollection.

src/DaemonRunner/DaemonRunner/Service/App/ICodeGenerationHandler.cs
helto4real: Not sure if it needs to be added to all async methods. Most are not long running.

helto4real (commented):
I think this is a great idea. this gives a good possibility to have several code generation methods.

src/DaemonRunner/DaemonRunner/Service/App/ICodeGenerationHandler.cs
Zorgino: offering only methods without a CT in an interface practically commits implementations to making the execution short, and if that's the case a reasonable question is why the method should be async at all.
adding CT to a method signature costs nothing, furthermore overloads without CT can be provided.
but in this scenario, we are providing an interface that can be potentially implemented with some long running logic. rendering our implication about short running task incorrect

helto4real (approved):

PR #379, opened 2021-05-23
More cleanup and fixing restart bug

helto4real:
## Proposed change

- Alot of more code smell fixed 
- continue break up NetDaemonHost
- Fixed restart bug
## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/NetDaemon.sln.DotSettings
+++ b/NetDaemon.sln.DotSettings
@@ -1,3 +1,0 @@
-﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
-	<s:Boolean x:Key="/Default/UserDictionary/Words/=deamon/@EntryIndexedValue">True</s:Boolean>
-	<s:Boolean x:Key="/Default/UserDictionary/Words/=hassio/@EntryIndexedValue">True</s:Boolean></wpf:ResourceDictionary>
--- a/NetDaemon.sln.DotSettings
+++ b/NetDaemon.sln.DotSettings
@@ -1,2 +1,3 @@
 ﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
-	<s:Boolean x:Key="/Default/UserDictionary/Words/=deamon/@EntryIndexedValue">True</s:Boolean></wpf:ResourceDictionary>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=deamon/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=hassio/@EntryIndexedValue">True</s:Boolean></wpf:ResourceDictionary>
--- a/NetDaemon.sln.DotSettings
+++ b/NetDaemon.sln.DotSettings
@@ -1,0 +1,2 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=deamon/@EntryIndexedValue">True</s:Boolean></wpf:ResourceDictionary>
--- a/src/Daemon/NetDaemon.Daemon/Daemon/Config/YamlAppConfig.cs
+++ b/src/Daemon/NetDaemon.Daemon/Daemon/Config/YamlAppConfig.cs
@@ -126,13 +126,13 @@ namespace NetDaemon.Daemon.Config
                     return null;
                 case YamlNodeType.Sequence:
                 {
-                    var list = InstanceSequence(deamonApp, parent, instanceType, node);
+                    var list = CreateSequenceInstance(deamonApp, parent, instanceType, node);
 
                     return list;
                 }
                 case YamlNodeType.Mapping:
                 {
-                    var instance = InstanceMapping(deamonApp, instanceType, node);
+                    var instance = CreateMappingInstance(deamonApp, instanceType, node);
 
                     return instance;
                 }
@@ -141,7 +141,7 @@ namespace NetDaemon.Daemon.Config
             }
         }
 
-        private object? InstanceMapping(INetDaemonAppBase deamonApp, Type instanceType, YamlNode node)
+        private object? CreateMappingInstance(INetDaemonAppBase deamonApp, Type instanceType, YamlNode node)
         {
             var instance = Activator.CreateInstance(instanceType);
 
@@ -183,7 +183,7 @@ namespace NetDaemon.Daemon.Config
         }
 
         [SuppressMessage("", "CA1508")]
-        private IList InstanceSequence(INetDaemonAppBase deamonApp, object? parent, Type instanceType, YamlNode node)
+        private IList CreateSequenceInstance(INetDaemonAppBase deamonApp, object? parent, Type instanceType, YamlNode node)
         {
             Type listType = instanceType?.GetGenericArguments()[0] ??
                             throw new NetDaemonNullReferenceException(
--- a/src/Daemon/NetDaemon.Daemon/Daemon/Config/YamlAppConfig.cs
+++ b/src/Daemon/NetDaemon.Daemon/Daemon/Config/YamlAppConfig.cs
@@ -113,78 +113,96 @@ namespace NetDaemon.Daemon.Config
         [SuppressMessage("", "CA1508")] // Weird bug that this should not warn!
         private object? InstanceProperty(INetDaemonAppBase deamonApp, object? parent, Type instanceType, YamlNode node)
         {
-            if (node.NodeType == YamlNodeType.Scalar)
+            switch (node.NodeType)
             {
-                var scalarNode = (YamlScalarNode) node;
-                ReplaceSecretIfExists(scalarNode);
-                return ((YamlScalarNode) node).ToObject(instanceType, deamonApp);
-            }
-            else if (node.NodeType == YamlNodeType.Sequence)
-            {
-                if (instanceType.IsGenericType && instanceType?.GetGenericTypeDefinition() == typeof(IEnumerable<>))
+                case YamlNodeType.Scalar:
                 {
-                    Type listType = instanceType?.GetGenericArguments()[0] ??
-                                    throw new NetDaemonNullReferenceException(
-                                        $"The property {instanceType?.Name} of Class {parent?.GetType().Name} is not compatible with configuration");
-
-                    IList list = listType.CreateListOfPropertyType() ??
-                                 throw new NetDaemonNullReferenceException(
-                                     "Failed to create listtype, please check {prop.Name} of Class {app.GetType().Name}");
-
-                    foreach (YamlNode item in ((YamlSequenceNode) node).Children)
-                    {
-                        var instance = InstanceProperty(deamonApp, null, listType, item) ??
-                                       throw new NotSupportedException(
-                                           $"The class {parent?.GetType().Name} has wrong type in items");
-
-                        list.Add(instance);
-                    }
+                    var scalarNode = (YamlScalarNode) node;
+                    ReplaceSecretIfExists(scalarNode);
+                    return ((YamlScalarNode) node).ToObject(instanceType, deamonApp);
+                }
+                case YamlNodeType.Sequence when !instanceType.IsGenericType ||
+                                                instanceType?.GetGenericTypeDefinition() != typeof(IEnumerable<>):
+                    return null;
+                case YamlNodeType.Sequence:
+                {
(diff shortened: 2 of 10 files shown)


NetDaemon.sln.DotSettings
```diff
@@ -0,0 +1,3 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=deamon/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=hassio/@EntryIndexedValue">True</s:Boolean></wpf:ResourceDictionary>
```
BeeHiveJava, line 3: Should this file be committed?

src/Daemon/NetDaemon.Daemon/Daemon/Config/YamlAppConfig.cs
```diff
@@ -113,78 +113,96 @@ public IEnumerable<INetDaemonAppBase> Instances
...
+
+                    return instance;
+                }
+                default:
+                    return null;
             }
-            else if (node.NodeType == YamlNodeType.Mapping)
+        }
+
+        private object? InstanceMapping(INetDaemonAppBase deamonApp, Type instanceType, YamlNode node)
```
BeeHiveJava, line 144: Can we put a verb in front of this? "CreateInstanceMapping"?

src/Daemon/NetDaemon.Daemon/Daemon/Config/YamlAppConfig.cs
```diff
@@ -113,78 +113,96 @@ public IEnumerable<INetDaemonAppBase> Instances
...
-                            break;
+            return instance;
+        }
 
-                        case YamlNodeType.Mapping:
-                            // Maps are not currently supported (var map = (YamlMappingNode)entry.Value;)
-                            break;
-                    }
+        [SuppressMessage("", "CA1508")]
+        private IList InstanceSequence(INetDaemonAppBase deamonApp, object? parent, Type instanceType, YamlNode node)
```
BeeHiveJava, line 186: Same thing here.

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
```diff
@@ -334,35 +337,26 @@ public async Task ReloadAllApps()
...
-            _cancelToken.ThrowIfCancellationRequested();
-
-            string? hassioToken = Environment.GetEnvironmentVariable("HASSIO_TOKEN");
-
-            if (_hassClient == null)
+            if (_hassClient is null)
             {
-                throw new NetDaemonNullReferenceException("Failed to instance HassClient!");
+                // If the host is reconnected the hass client will be null so lets instance new one
+                _hassClient = _hassClientFactory.New() ??
```
FrankBakkerNl, line 343: Shouldn't the client always be recreated in the Run method? I thought that is how it was originalky

NetDaemon.sln.DotSettings
helto4real: It was kinda annoying that it was complaining about spelling errors of things that will never be fixed. We could remove it too. It is not a deal breaker.

src/Daemon/NetDaemon.Daemon/Daemon/Config/YamlAppConfig.cs
helto4real: Yes! But it is an verb. Instance. But I agree I could make it even more clear

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
helto4real: Yea  good point

src/Daemon/NetDaemon.Daemon/Daemon/Config/YamlAppConfig.cs
BeeHiveJava: Oh that's okay. But then it should be "InstantiateMapping". I think "CreateMapping" and "CreateSequence" are a little simpler though.

helto4real: Agree will fix

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
```diff
@@ -436,12 +410,67 @@ public async Task Run(string host, short port, bool ssl, string token, Cancellat
                 // Set cancel token to avoid background processes
                 // to access disconnected Home Assistant
                 IsConnected = false;
-                _cancelTokenSource.Cancel();
+                if (_cancelTokenSource != null)
```
BeeHiveJava, line 413: Invert this if-statement to reduce nesting.

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
```diff
@@ -436,12 +410,67 @@ public async Task Run(string host, short port, bool ssl, string token, Cancellat
                 // Set cancel token to avoid background processes
                 // to access disconnected Home Assistant
                 IsConnected = false;
-                _cancelTokenSource.Cancel();
+                if (_cancelTokenSource != null)
+                {
+                    _cancelTokenSource.Cancel();
+                    try
+                    {
+                        Task.WaitAll(runningTasks.ToArray(), TimeSpan.FromSeconds(1));
```
BeeHiveJava, line 418: Can we use Task.WhenAll (async variant) and await that, or os that not possible here?

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
```diff
@@ -436,12 +410,67 @@ public async Task Run(string host, short port, bool ssl, string token, Cancellat
...
+                {
+                    _cancelTokenSource.Cancel();
+                    try
+                    {
+                        Task.WaitAll(runningTasks.ToArray(), TimeSpan.FromSeconds(1));
+                    }
+                    catch (Exception e)
+                    {
+                        Logger.LogTrace(e, "Failed to cancel the running tasks");
+                        // We do not handle error here
```
BeeHiveJava, line 423: Why not? (this comment doesn't say much. I can see that we don't, but I don't know the reason why)

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
```diff
@@ -436,12 +410,67 @@ public async Task Run(string host, short port, bool ssl, string token, Cancellat
...
+        /// </summary>
+        /// <param name="cancellationToken">Token to cancel current operation</param>
+        /// <returns>true if read operation is success</returns>
+        /// <exception cref="NetDaemonNullReferenceException"></exception>
+        private async Task<bool> ReadEvent()
+        {
+            _ = _hassClient ?? throw new NetDaemonNullReferenceException("Failed to instance HassClient!");
+
+            HassEvent changedEvent = await _hassClient.ReadEventAsync(_cancelToken).ConfigureAwait(false);
+            if (changedEvent != null)
```
BeeHiveJava, line 440: Can we invert this if-statement?

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
```diff
@@ -436,12 +410,67 @@ public async Task Run(string host, short port, bool ssl, string token, Cancellat
...
+                if (changedEvent.Data is HassServiceEventData hseData && hseData.Domain == "homeassistant" &&
+                    (hseData.Service == "stop" || hseData.Service == "restart"))
+                {
+                    // The user stopped HA so just stop processing messages
+                    Logger.LogInformation("User {action} Home Assistant, will try to reconnect...",
+                        hseData.Service == "stop" ? "stopping" : "restarting");
+                    return false;
+                }
+
+                // Remove all completed Tasks
```
BeeHiveJava, line 451: This comment adds nothing (code is clear enough).

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
```diff
@@ -529,7 +552,7 @@ public async Task SetDaemonStateAsync(int numberOfLoadedApps, int numberOfRunnin
             }
             else
             {
-                return  SetStateAndWaitForResponseAsync(entityId, state, attributes, true).Result;
+                return SetStateAndWaitForResponseAsync(entityId, state, attributes, true).Result;
```
BeeHiveJava, line 555: Looking at this now, I don't think we should support any kind of scenario where call an async method on a sync context.

BeeHiveJava: Not that we have to fix this now.

src/Daemon/NetDaemon.Daemon/Daemon/Config/YamlExtensions.cs
```diff
@@ -45,13 +43,11 @@ public static class YamlExtensions
...
-            Type? underlyingNullableType = Nullable.GetUnderlyingType(valueType);
+            var underlyingNullableType = Nullable.GetUnderlyingType(valueType);
 
             if (underlyingNullableType != null)
-            {
                 // It is nullable type
                 valueType = underlyingNullableType;
-            }
 
             switch (valueType.Name)
```
FrankBakkerNl, line 52: I think this could be a nice switch expression, maybe with 'when' clauses

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
helto4real: Yes good catch!

src/Daemon/NetDaemon.Daemon/Daemon/Config/YamlExtensions.cs
FrankBakkerNl, line 52: Maybe for another day, but switching on the type would feel better then switch in the Type.Name

BeeHiveJava: Good one.

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
```diff
@@ -92,6 +92,10 @@ public class NetDaemonHost : INetDaemonHost, IAsyncDisposable
 
         internal bool HasNetDaemonIntegration;
 
+        // This is non null if this is running in a add-on
+        private readonly string? _addOnToken = Environment.GetEnvironmentVariable("HASSIO_TOKEN");
+        private readonly IHassClientFactory _hassClientFactory;
+        public bool IsAddOn => _addOnToken != null;
```
FrankBakkerNl, line 98: Is this new or moved? mabe also make it internal as not to expand the public API

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
helto4real: That should be another PR. Not sure how to solve it though

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
helto4real: yes should be internal.

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
BeeHiveJava: Agreed. Definitely not for now.

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
FrankBakkerNl: Maybe the tasks could self-remove from the list.
If when adding them to the list we do something like:

task.ContinueWith(t => _eventHandlerTasks.Remove(t)
 We no longer need to check all tasks everytime.

Also I think this List should be a ConcurrentList as items are added and removed from possible multiple threads

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
helto4real: actually I added logging there so the comment is left over

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
FrankBakkerNl: We will have to clean this API fr sure we have 3 different public SetState... methods, all with slightly different behaviour

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
helto4real: cannot jump out of a finally block

helto4real: I found a way to remove it

src/Daemon/NetDaemon.Daemon/Daemon/Config/YamlExtensions.cs
helto4real: But the switch case would be case typeof(int) and I thought that looked even worse

helto4real: Maybe, I tried but did not come up with anything that looked better.

helto4real:
@BeeHiveJava @FrankBakkerNl thanks for the great comments on the review. I fixed most of them now. If anyone of you think it is ready to merge approve and merge please.

FrankBakkerNl (approved):

PR #421, opened 2021-07-31
Di on application

FrankBakkerNl:
## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist


src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
```diff
@@ -696,34 +696,29 @@ internal async Task RefreshInternalStatesAndSetArea()
...
-                        {
-                            throw new NetDaemonException(
-                                $"There is no app named {dependency}, please check dependencies or make sure you have not disabled the dependent app!");
-                        }
+            if (dependencies.Count == 0) return unsortedList;
 
-                        edges.Add(new Tuple<INetDaemonAppBase, INetDaemonAppBase>(instance, dependentApp));
-                    }
-                }
+            // just make sure we  have no null id's
```
FrankBakkerNl, line 705: Now the app doe sno longer ned to implement IEqualityComparere which simplifies things some more
Need to make some more tests for this, I think it is still correct but did quite some refactoring.

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
```diff
@@ -1140,34 +1139,43 @@ private async Task SetDependentState(string entityId, string state)
         }
 
         //TODO: Refactor this
-        private async Task PersistAppStateAsync(NetDaemonAppBase app)
+        private async Task PersistAppStateAsync(INetDaemonApp app)
         {
-            var obj = await GetDataAsync<IDictionary<string, object?>>(app.GetUniqueIdForStorage())
+            var uniqueIdForStorage = $"{app.GetType().Name}_{app.Id}".ToLowerInvariant();
```
FrankBakkerNl, line 1144: I copied the code from GetUniqueIdForStorage() to here so it would not have to be on the interface. I wonder however what exactly this code does, i cannot find where this persisted data is read

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
```diff
@@ -1140,34 +1139,43 @@ private async Task SetDependentState(string entityId, string state)
...
-        private async Task<bool> RestoreAppState(INetDaemonAppBase appInstance)
+        private async Task<bool> RestoreAppState(INetDaemonApp appInstance)
         {
             try
             {
                 // First do startup initialization to connect with this daemon instance
-                await appInstance.StartUpAsync(this).ConfigureAwait(false);
+                if (appInstance is INetDaemonAppBase netDaemonAppBase)
+                {
+                    await netDaemonAppBase.StartUpAsync(this).ConfigureAwait(false);
```
FrankBakkerNl, line 1161: StartUpAsync Is not needed for 'new' apps to get the reference to the NetDaemonHost because they can get the dependencies via the ctor. It looks like there is allways a single NetDaemonHost anyways so there is no need to provide it again when the app restarts.

Maybe we do need a Start / Stop method on the INetDaemonApp interface, or maybe on a separate interface so an app can opt-in for start stop events

src/DaemonRunner/DaemonRunner/Infrastructure/Extensions/AssemblyExtensions.cs
```diff
@@ -10,7 +10,7 @@ internal static class AssemblyExtensions
         public static IEnumerable<Type> GetTypesWhereSubclassOf<T>(this Assembly assembly)
         {
             return assembly.GetTypes()
-                .Where(type => type.IsClass && !type.IsGenericType && !type.IsAbstract && type.IsSubclassOf(typeof(T)));
+                .Where(type => type.IsClass && !type.IsGenericType && !type.IsAbstract && type.IsAssignableTo(typeof(T)));
```
FrankBakkerNl, line 13: This makes sure we can load any class that implements INetDaemonApp without inheriting a base class
(I see now the name GetTypesWhereSubclassOf() no longer applies)

src/App/NetDaemon.App/Common/NetDaemonAppBase.cs
```diff
@@ -106,15 +106,6 @@ public string Description
 
         internal FluentExpandoObject? InternalStorageObject { get; set; }
 
-        /// <summary>
```
FrankBakkerNl, line 109: Looks like this was only used for the dependency sorting whuch uses just the ids now

src/App/NetDaemon.App/Common/INetDaemonAppLogging.cs
```diff
@@ -0,0 +1,158 @@
+﻿using System;
+
+namespace NetDaemon.Common
+{
+    /// <summary>
+    ///     Interface for logging capabilities in NetDaemon Apps
+    /// </summary>
+    public interface INetDaemonAppLogging
```
BeeHiveJava, line 8: By the way, why don't we just let this implement ILogger? NetDaemon already depends on the Microsoft Logging framework anyway.

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
```diff
@@ -563,15 +562,16 @@ public async Task UnloadAllApps()
         {
             Logger.LogTrace("Unloading all apps ({instances}, {running})", InternalAllAppInstances.Count,
                 InternalRunningAppInstances.Count);
-            foreach (var app in InternalAllAppInstances)
+            foreach (var app in InternalAllAppInstances.Values)
             {
                 try
                 {
-                    await app.Value.DisposeAsync().ConfigureAwait(false);
+                    if (app is IAsyncDisposable asyncDisposable)
```
BeeHiveJava, line 569: Should we also check / dispose for regular disposable here?

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
```diff
@@ -991,7 +986,7 @@ private async Task LoadAllApps()
                 InternalRunningAppInstances.Clear();
             }
 
-            foreach (INetDaemonAppBase appInstance in instancedApps!)
+            foreach (INetDaemonApp appInstance in instancedApps!)
```
BeeHiveJava, line 989: Can we use var here?

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
BeeHiveJava: Yes. The important thing to consider though is the lifetime of apps. Apps can be become singletons (which may or may not be okay) if they're never disposed of.

I think everything works as expected, it's just something to consider if we ever end up modifying the lifetime of either the host or the apps.

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
```diff
@@ -1140,34 +1139,43 @@ private async Task SetDependentState(string entityId, string state)
...
+                if (appInstance is INeatDaemonPersistantApp persistantApp)
+                {
+                    await persistantApp.RestoreAppStateAsync().ConfigureAwait(false);
+                }
 
-                if (!appInstance.IsEnabled)
+                if (appInstance.IsEnabled) return true;
+                
+                // We should not initialize this app, so dispose it and return
+                if (appInstance is IAsyncDisposable asyncDisposable)
```
BeeHiveJava, line 1173: Also here, should we check / dispose for regular disposable?

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
FrankBakkerNl: Yes that would make sense

src/App/NetDaemon.App/Common/INetDaemonAppLogging.cs
FrankBakkerNl: I think we don't need this interface anymore but need to keep it for backwards comoatibility

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
FrankBakkerNl: Yea think so

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
FrankBakkerNl: If Inderstand correct each app is created onky once and stopped / started without creating a new instance. Maybe creating a new instance each time it starts will make rhinhs easier but this will be very much a breaking change.

src/App/NetDaemon.App/Common/INetDaemonAppLogging.cs
helto4real: The logging is convenience functions. But in retroperspective I could just been exposing the ILogger. So keep for backwards compat.

src/App/NetDaemon.App/Common/INetDaemonInitialableApp.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System.Threading.Tasks;
+
+namespace NetDaemon.Common
+{
+    /// <summary>
+    ///     Interface for applications that support Saving nd restoring their state
+    /// </summary>
+    public interface INeatDaemonPersistantApp
```
helto4real, line 8: The classname or filename shoudl change name?

src/App/NetDaemon.App/Common/NetDaemonAppBase.cs
```diff
@@ -299,7 +290,7 @@ public async virtual ValueTask DisposeAsync()
         }
 
         /// <inheritdoc/>
-        public INetDaemonAppBase? GetApp(string appInstanceId)
+        public INetDaemonApp? GetApp(string appInstanceId)
```
helto4real, line 293: Like the name change. Is better.

src/Daemon/NetDaemon.Daemon/Daemon/Config/YamlAppConfig.cs
```diff
@@ -69,15 +73,15 @@ public IEnumerable<INetDaemonAppBase> GetInstances()
...
+        public INetDaemonApp? InstanceAndSetPropertyConfig(
             Type netDaemonAppType,
             YamlMappingNode appNode,
             string? appId)
         {
             _ = appNode ??
                 throw new NetDaemonArgumentNullException(nameof(appNode));
 
-            var netDaemonApp = (INetDaemonAppBase?) Activator.CreateInstance(netDaemonAppType);
+            var netDaemonApp = (INetDaemonApp)ActivatorUtilities.CreateInstance(_serviceProvider, netDaemonAppType);
```
helto4real, line 84: Cool the new ActivatorUtilities!

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
helto4real: Yea this needs tests. I spent alot of manual time getting this right.  But automatic tests would be better ofc.

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
helto4real: This is for the `Store` feature. If you use `Store.any_kind_of_thing = "hello"`it will store it as a key/value using app id to make it unique for app

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
helto4real: No they are removed. Everytime an app is disabled all apps Unloads and Loads again so the dependencies is right. Letś say you disable an app that another app is depending on that app will be disabled too. Maybe we could come up with another smarter way of app lifecycle. See the `SetStateOnDaemonAppSwitch` function. It calls `ReloadAllApps()`

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
helto4real: Not sure why this is needed. The apps does not implement the IDisposable. But it is no harm ether.

src/DevelopmentApps/apps/InterfaceApp/InterfaceApp.cs
```diff
@@ -0,0 +1,40 @@
...
+using Microsoft.Extensions.Logging;
+using NetDaemon.Common;
+using NetDaemon.Daemon;
+
+namespace NetDaemon.DevelopmentApps.apps.DebugApp
+{
+    /// <summary> Test application for interface based app
+    /// </summary>
+    [Focus]
+    public class InterfaceApp : INetDaemonApp
```
helto4real, line 15: This looks really good!

src/App/NetDaemon.App/Common/NetDaemonAppBase.cs
FrankBakkerNl, line 293: This would theoretically be a breaking change is this is used by a client app

src/App/NetDaemon.App/Common/INeatDaemonPersistantApp.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System.Threading.Tasks;
+
+namespace NetDaemon.Common
+{
+    /// <summary>
+    ///     Interface for applications that support Saving nd restoring their state
+    /// </summary>
+    public interface INeatDaemonPersistantApp
```
Zorgino, line 8: is Neat intended?

PR #422, opened 2021-08-06
Allow classes without base class to be loaded as app

FrankBakkerNl:
## Breaking change

TODO: Check what exactly is broken, maybe some of it can be undone and list the rest here

## Proposed change

Enabled any class to be used as a NetDaemon app by using the [NetDaemonApp] attribute and not requiring a specific baseclass or interface.

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/src/App/NetDaemon.App/Common/ApplicationContext.cs
+++ b/src/App/NetDaemon.App/Common/ApplicationContext.cs
@@ -1,0 +1,102 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Reflection;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace NetDaemon.Common
+{
+    internal class ApplicationMetaData : IApplicationMetadata
+    {
+        public string? Id { get; set; }
+        public string? Description { get; set; } = null;
+        public bool IsEnabled { get; set; } = true;
+        public AppRuntimeInfo RuntimeInfo { get; } = new AppRuntimeInfo();
+    }
+
+    public class ApplicationContext
+    {
+        private readonly ILogger _logger;
+        private IApplicationMetadata _applicationMetadata;
+
+        public ApplicationContext(object applicationInstance, ILogger logger)
+        {
+            // For old style applications the metadata properties are stored in the app itself
+            // we need to keep that for backwards compatibility, for new apps we will use a separate class for the
+            // metadata properties
+            _applicationMetadata = applicationInstance is IApplicationMetadata appAsMetaData
+                ? appAsMetaData
+                : new ApplicationMetaData();
+          
+            ApplicationInstance = applicationInstance;
+            _logger = logger;
+        }
+        public ApplicationContext(IAsyncInitializable netDaemonApp) : this(netDaemonApp, NullLogger.Instance)
+        { }
+        public object ApplicationInstance { get; }
+        
+        /// <summary>
+        ///     Unique id of the application
+        /// </summary>
+        public string? Id
+        {
+            get => _applicationMetadata.Id;
+            init => _applicationMetadata.Id = value;
+        }
+
+        /// <summary>
+        ///     The dependencies that needs to be initialized before this app
+        /// </summary>
+        public IEnumerable<string> Dependencies { get; set; } = Array.Empty<string>();
+
+        public string? Description => _applicationMetadata.Description
+                                      ?? ApplicationInstance.GetType().GetCustomAttribute<DescriptionAttribute>()?.Description 
+                                      ?? "";
+
+        /// <summary>
+        ///     Gets or sets a flag indicating whether this app is enabled.
+        ///     This property can be controlled from Home Assistant.
+        /// </summary>
+        /// <remarks>
+        ///     A disabled app will not be initialized during the discovery.
+        /// </remarks>
+        public bool IsEnabled
+        {
+            get => _applicationMetadata.IsEnabled;
+            set => _applicationMetadata.IsEnabled = value;
+        }
+
+        /// <summary>
+        ///     Returns different runtime information about an app
+        /// </summary>
+        public AppRuntimeInfo RuntimeInfo => _applicationMetadata.RuntimeInfo;
+
+        /// <summary>
+        ///     Unique id of the application entity
+        /// </summary>
+        public string EntityId => _applicationMetadata.EntityId;
+
+
+        public async Task UnloadAsync()
+        {
+            try
+            {
+                if (ApplicationInstance is IAsyncDisposable asyncDisposable)
+                {
+                    await asyncDisposable.DisposeAsync().ConfigureAwait(false);
+                }
+
+                if (ApplicationInstance is IDisposable disposable)
+                {
+                    disposable.Dispose();
+                }
+            }
+            catch (Exception e)
+            {
+                _logger.LogError(e, "Failed to unload apps, {app_id}", Id);
+            }
+        }
+    }
+}
--- a/src/App/NetDaemon.App/Common/ApplicationContext.cs
+++ b/src/App/NetDaemon.App/Common/ApplicationContext.cs
@@ -2,22 +2,12 @@
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Reflection;
-using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Logging.Abstractions;
 using NetDaemon.Daemon.Services;
 
 namespace NetDaemon.Common
 {
-    internal class ApplicationMetaData : IApplicationMetadata
-    {
-        public string? Id { get; set; }
-        public string? Description { get; set; } = null;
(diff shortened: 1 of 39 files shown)


src/App/NetDaemon.App/Common/ApplicationContext.cs
```diff
@@ -0,0 +1,110 @@
...
+{
+    internal class ApplicationMetaData : IApplicationMetadata
+    {
+        public string? Id { get; set; }
+        public string? Description { get; set; } = null;
+        public bool IsEnabled { get; set; } = true;
+        public AppRuntimeInfo RuntimeInfo { get; } = new AppRuntimeInfo();
+    }
+
+    public class ApplicationContext
```
FrankBakkerNl, line 21: This class now wraps the actual app instance. In many places where I used INetDaemonApp befor I now use this (INetDaemonApp is actually removed)

src/App/NetDaemon.App/Common/ApplicationContext.cs
```diff
@@ -0,0 +1,110 @@
...
+    {
+        private readonly ILogger _logger;
+        private IApplicationMetadata _applicationMetadata;
+        private INetDaemonPersistantApp _persistantApp;
```
FrankBakkerNl, line 25: Not used yet, but to try how to make features from the original AppBase class reuable to classes that do not derive from AppBase

src/App/NetDaemon.App/Common/ApplicationPersistenceService.cs
```diff
@@ -0,0 +1,112 @@
...
+using System.Threading;
+using System.Threading.Channels;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using NetDaemon.Common;
+using NetDaemon.Common.Exceptions;
+
+namespace NetDaemon.Daemon.Services
+{
+    internal class ApplicationPersistenceService : INetDaemonPersistantApp
```
FrankBakkerNl, line 14: This is basically the RestoreAppStateAsync and SaveAppState extarcted from the AppBase

src/App/NetDaemon.App/Common/INetDaemon.cs
```diff
@@ -48,7 +48,7 @@ public interface INetDaemon : INetDaemonCommon
         ///     Get application instance by application instance id
         /// </summary>
         /// <param name="appInstanceId">The identity of the app instance</param>
-        INetDaemonApp? GetApp(string appInstanceId);
+        INetDaemonAppBase? GetApp(string appInstanceId);
```
FrankBakkerNl, line 51: Reverted back to what it was

src/App/NetDaemon.App/Common/ApplicationContext.cs
```diff
@@ -0,0 +1,110 @@
...
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+using NetDaemon.Daemon.Services;
+
+namespace NetDaemon.Common
+{
+    internal class ApplicationMetaData : IApplicationMetadata
```
Zorgino, line 13: change MetaData to Metadata, please

src/App/NetDaemon.App/Common/INetDaemonInitialableApp.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System.Threading.Tasks;
+
+namespace NetDaemon.Common
+{
+    /// <summary>
+    ///     Apps that can be initialized
+    /// </summary>
+    public interface INetDaemonInitialableApp
```
Zorgino, line 8: please, change Initialable to Initializable

Zorgino: and, maybe, compose this interface from IInitializable, IAsyncInitializible and INetDaemonPersistantApp?

tests/NetDaemon.Daemon.Tests/DaemonRunner/Config/ConfigTest.cs
```diff
@@ -279,8 +281,8 @@ public void YamlAdvancedObjectsShouldReturnCorrectData()
             yamlStream.Load(new StringReader(yaml));
             var root = (YamlMappingNode) yamlStream.Documents[0].RootNode;
 
-            var instance =
-                (AppComplexConfig?) yamlConfig.InstanceAndSetPropertyConfig(typeof(AppComplexConfig), root, "id");
+            var instance = new AppComplexConfig(); 
+            yamlConfig.SetPropertyConfig(root, "id", null!, instance);
```
Zorgino, line 285: null! - how that even works, lol

Zorgino:
outstanding pr, will definitely check it out after merge. well done!

src/App/NetDaemon.App/Common/ApplicationContext.cs
FrankBakkerNl: Fixed + moved to other file

src/App/NetDaemon.App/Common/INetDaemonInitialableApp.cs
FrankBakkerNl: This typo was actually already in the current dev branch. I tried to undo some of the breaking changes I did in the previous PR (unfortunately fixin the spelling is a breaking change). I doubt is anyone actually uses this interface. There are a lot of public  types and methods that I think are probably not used by anyone. For now I will try to keep it compatible as much as possible and maby we can later mark them [Obsolete}

src/App/NetDaemon.App/Common/ApplicationContext.cs
```diff
@@ -0,0 +1,100 @@
...
+
+namespace NetDaemon.Common
+{
+    public class ApplicationContext
+    {
+        private readonly ILogger _logger;
+        private IApplicationMetadata _applicationMetadata;
+        private INetDaemonPersistantApp _persistantApp;
+
+        public ApplicationContext(object applicationInstance, INetDaemon netDaemon, ILogger logger)
```
BeeHiveJava, line 17: I'm just really questioning this pattern. Why do we need the application instance? And if we really need it, why don't we make an interface `IApplicationInstance` (or something) that exposes a `IApplicationMetadata` and `INetDaemonPersistantApp`? I think that's a lot cleaner than passing an object + it explains intent.

src/App/NetDaemon.App/Common/ApplicationPersistenceService.cs
helto4real: Nice!

src/App/NetDaemon.App/Common/ApplicationContext.cs
helto4real: Seems like a good path forward

src/App/NetDaemon.App/Common/INetDaemonInitialableApp.cs
helto4real: I think it is fine to rename it

tests/NetDaemon.Daemon.Tests/DaemonRunner/Config/ConfigTest.cs
helto4real: I did not ether hehe

src/App/NetDaemon.App/Common/ApplicationContext.cs
helto4real: Yea interface would be better and also makes tests easier to write (unless I am missing something here)

helto4real:
I like where this is headed overall. Good job. Please rename the PR to something a bit more description since the heading and also the PR description is auto generated release notes later on.

src/App/NetDaemon.App/Common/ApplicationContext.cs
FrankBakkerNl: What I more or less tried to do is replace inheritance with containment. The application context is now responsible for handling the application.  So it needs the instance of the app to do that,.

The challenging part was to keep the old model working as well where the app itself has the public methods that are now moved to the ApplicationContext.

Another option might be to have the ApplicationContext create the instance of the app but I think that might be harder to do in a backwards compatible way. But maybe I wil try to do that in a next PR because it might make other things easier. It will at least avoid passing an object in. The other two ctor arguments here I think will also be replaced by tan IServiceProvider eventually

The reason for this is to allow apps to be as simple as possible which allows different kind of models to be used. An app does not even need to know anything about NetDaemon or HomeAssistant if it does not want to, but you would be able to start and stop it from HA.

First I did this with the INetDaemonApp interface which had some properties and InitAsync method. But now those are not really mandadory anymore (eg, you can implement IInitializable  or IAsyncInitializable dependng n your needs) It feels kind of weird to have a 'marker' interface without any members on it.

helto4real: Ahh right. Yea was thinking that too an emtpy interface but in that case an object seem more reasonable.

helto4real: @BeeHiveJava what you think?

helto4real (approved):

PR #428, opened 2021-08-14
move app instantiation from YamlAppConfig.cs to a new class

Zorgino:
## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/src/Daemon/NetDaemon.Daemon/Daemon/AppInstantiator.cs
+++ b/src/Daemon/NetDaemon.Daemon/Daemon/AppInstantiator.cs
@@ -1,0 +1,34 @@
+﻿using System;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using NetDaemon.Common;
+
+namespace NetDaemon.Daemon
+{
+    public class AppInstantiator : IAppInstantiator
+    {
+        private readonly IServiceProvider _serviceProvider;
+        private readonly INetDaemon _netDaemon;
+        private readonly ILogger _logger;
+
+        public AppInstantiator(IServiceProvider serviceProvider)
+        {
+            _serviceProvider = serviceProvider;
+            
+            _netDaemon = _serviceProvider.GetRequiredService<INetDaemon>();
+            _logger = _serviceProvider.GetRequiredService<ILogger>();
+        }
+        
+        public ApplicationContext Instantiate(Type netDaemonAppType, string appId)
+        {
+            var appInstance = ActivatorUtilities.CreateInstance(_serviceProvider, netDaemonAppType);
+
+            var appContext = new ApplicationContext(appInstance, _netDaemon, _logger)
+            {
+                Id = appId
+            };
+            
+            return appContext;
+        }
+    }
+}
--- a/src/Daemon/NetDaemon.Daemon/Daemon/AppInstantiator.cs
+++ b/src/Daemon/NetDaemon.Daemon/Daemon/AppInstantiator.cs
@@ -5,7 +5,7 @@ using NetDaemon.Common;
 
 namespace NetDaemon.Daemon
 {
-    public class AppInstantiator : IAppInstantiator
+    internal class AppInstantiator : IAppInstantiator
     {
         private readonly IServiceProvider _serviceProvider;
         private readonly INetDaemon _netDaemon;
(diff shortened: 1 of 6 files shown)


src/Daemon/NetDaemon.Daemon/Daemon/AppInstantiator.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using NetDaemon.Common;
+
+namespace NetDaemon.Daemon
+{
+    public class AppInstantiator : IAppInstantiator
```
FrankBakkerNl, line 8: Can this be internal to limit the public API surface?

src/Daemon/NetDaemon.Daemon/Daemon/Config/YamlAppConfig.cs
```diff
@@ -73,24 +71,18 @@ public IEnumerable<ApplicationContext> GetInstances()
...
 
-            var appInstance = ActivatorUtilities.CreateInstance(_serviceProvider, netDaemonAppType);
-            
-            var appContext = new ApplicationContext(appInstance, 
-                _serviceProvider.GetRequiredService<INetDaemon>(),
-                _serviceProvider.GetRequiredService<ILogger>())
-            {
-                Id = appId
-            };
+            var appContext = _appInstantiator.Instantiate(netDaemonAppType, appId);
```
FrankBakkerNl, line 82: Maybe we should not do this from the yamlAppConfig eigther, but move it out. Not sure if that is easy bcause here you know which classes to instantiate

src/Daemon/NetDaemon.Daemon/Daemon/IAppInstantiator.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+using NetDaemon.Common;
+
+namespace NetDaemon.Daemon
+{
+    public interface IAppInstantiator
+    {
```
FrankBakkerNl, line 7: Does this need an interface?

src/Daemon/NetDaemon.Daemon/Daemon/Config/YamlAppConfig.cs
Zorgino: yes, we should definitely move it out from there. but goal is to make PRs as atomic as possible

src/Daemon/NetDaemon.Daemon/Daemon/IAppInstantiator.cs
Zorgino: "program to interfaces, not implementations"

FrankBakkerNl: Personally I think that is like a religious quote that is used way to often.  .Net classes do a very nice job of hiding implementation details behind public members without needing to duplicate them in a separate interface.

helto4real: I think it is fine. You think the PR could be merged @FrankBakkerNl ?

Zorgino: > Personally I think that is like a religious quote that is used way to often. .Net classes do a very nice job of hiding implementation details behind public members without needing to duplicate them in a separate interface.

in general, agreed. but without an interface testing is hard

FrankBakkerNl: Don't see why this cannot be tested without an interface. Interfaces help swapping in different implementations like mocks, but it is not likely you will mock this class. When the need arrives creating an interface then will be easy

But this is not something ill make a big fuzz about, maybe just make the interface internal as well

Zorgino: for now it's too basic for unit testing, it's true. but it is likely to acquire some extra logic. and it will be too late to create an interface without facing extra risks.

if it is a part of public contract of another class, in this case constructor of YamlConfig (DI eliminates such contracts as a class and I would readily use it instead of direct instantiation, but nevertheless), then any change to a contract violates open closed principle and leads to a snowball refactoring.

Zorgino: > Don't see why this cannot be tested without an interface. Interfaces help swapping in different implementations like mocks, but it is not likely you will mock this class. When the need arrives creating an interface then will be easy
> 
> But this is not something ill make a big fuzz about, maybe just make the interface internal as well

it won't compile as it is an arg of a public constructor in a public class YamlConfig.
but I generally like the idea to hide all stuff that should be internal

PR #434, opened 2021-08-19
M3 on remodel

FrankBakkerNl:
## Proposed change

Rebased and adjusted Model3 API to the remodel branch

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist


Model3/ModelV3/Internal/HaContextProvider.cs
```diff
@@ -0,0 +1,58 @@
...
+namespace NetDaemon.Common.ModelV3
+{
+    internal class HaContextProvider : IHaContext
+    {
+        private readonly NetDaemonHost _netDaemonHost;
+        private readonly ObservableBase<StateChange> _observable;
+        
+        // TODO: Initialize the stateCache here, would need some sort of init event, and access to the IHassClient 
+        private readonly EntityStateCache _entityStateCache = new ();
+        public HaContextProvider(NetDaemonHost netDaemonHost)
```
eugeneniemand, line 21: Should this not be INetDaemonHost? Just thinking about DI and Testing

Model3/ModelV3/Internal/HaContextProvider.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                observer.OnNext(stateChangedEventData.Map(this));
+            }
+        }
+
+        public EntityState? GetState(string entityId)
+        {
+            var hassState = _entityStateCache.GetState(entityId);
+            
+            return HassObjectMapperMapper.Map(hassState);
```
eugeneniemand, line 46: MapperMapper.Map 🤣 read that 10 times quickly

Model3/DependencySetup.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        public static IHostBuilder UseNetDaemonV3(this IHostBuilder hostBuilder)
+        {
+            if (hostBuilder == null) throw new ArgumentNullException(nameof(hostBuilder));
+
+            return hostBuilder
+                .ConfigureServices((_, services) =>
+                {
+                    // TODO: make this scoped and just the StateManager as a singleton
+                    services.TryAddSingleton<IHaContext>(provider => new HaContextProvider(provider.GetRequiredService<NetDaemonHost>()));
```
BeeHiveJava, line 20: Alternatively you can add HaContextProvider as Transient and then get that from the dependency container here. That way you don't have to inject dependencies into HaContextProvider yourself.

Model3/ModelV3/Domains/ZoneEntity.cs
```diff
@@ -0,0 +1,27 @@
...
+        public ZoneState(EntityState source) : base(source) { }
+    }
+    
+    public record ZoneAttributes
+    {
+        public bool hidden { get; init; }
+        public double latitude { get; init; }
+        public double longitude { get; init; }
+        public double radius { get; init; }
+        public string? friendly_name { get; init; }
```
BeeHiveJava, line 24: Should these be CamelCase with a json attribute instead?

src/Daemon/NetDaemon.Daemon/Daemon/NetDaemonHost.cs
```diff
@@ -123,6 +122,9 @@ internal void AddRunningApp(INetDaemonAppBase app)
             InternalRunningAppInstances[applicationContext.Id!] = applicationContext;
             InternalAllAppInstances[applicationContext.Id!] = applicationContext;
         }
+
+        // TODO: maybe find a nicer way to get events from here to the model3 API
```
BeeHiveJava, line 126: MediatR?

Model3/ModelV3/Internal/HaContextProvider.cs
FrankBakkerNl: Yes that would be better, the problem was that it is not registered as the intetface but we could  change that ofcourse

Model3/ModelV3/Internal/HassObjectMapperMapper.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            if (hassState == null) return null;
+            
+            return new EntityState()
+            {                
+                EntityId = hassState.EntityId,
+                State = hassState.State,
+                AttributesJson = hassState.AttributesJson ?? new JsonElement(),
+                LastChanged = hassState.LastChanged,
+                LastUpdated = hassState.LastUpdated,
```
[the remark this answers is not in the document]
FrankBakkerNl: These types map 1:1 to the HA structure (or at least they should) so I would like to keep them as clean as possible. If we need extra metadata about the status in the cache we could add them on a separate class maybe, but I am not sure what data exacly you mean.

Model3/ModelV3/EntityStateCache.cs
```diff
@@ -0,0 +1,28 @@
...
+                _latestStates[state.EntityId] = state;
+            }
+        }
+
+        
+        private ConcurrentDictionary<string, HassState> _latestStates = new ();
+
+        public void Store(HassState newState) => _latestStates[newState.EntityId] = newState;
+
+        public HassState? GetState(string entityId) => _latestStates.TryGetValue(entityId, out var state) ? state : null;
```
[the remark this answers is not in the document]
FrankBakkerNl: I think all entities that exist should have a state., at least I do not know how it coulds have no state unless there is a bug or other error situation.

 If that indeed is true null would mean the entity does not exist.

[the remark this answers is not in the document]
FrankBakkerNl: I was also thinking maybe we should lazy load the cache only when a state is requested or changed instead of prefetching all states

Model3/ModelV3/Internal/HassObjectMapperMapper.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            return new StateChange(
+                new Entity(haContext, source.EntityId),
+                Map(source.OldState),
+                Map(source.NewState));
+        }
+
+        public static EntityState? Map(HassState? hassState)
+        {
+            if (hassState == null) return null;
```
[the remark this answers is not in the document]
FrankBakkerNl: The state change New and Old properties are nullable in hassClient. For Old this makes sense buy for New a am not sure if it can actually be null, it would be strange if it is. In any case for the Old we need to map null -> null

Model3/DependencySetup.cs
FrankBakkerNl: Ah yes this is a bit messy indeed that indeed sounds better.

Still need to figure out what exactly should be scoped per app and what shouls be shared. Maybe need to introduce some more classes to seoarate things

Model3/ModelV3/Domains/ZoneEntity.cs
FrankBakkerNl: Added a todo for now

Model3/ModelV3/EntityState.cs
```diff
@@ -0,0 +1,37 @@
...
+using JoySoftware.HomeAssistant.Model;
+using Model3;
+
+namespace NetDaemon.Common.ModelV3
+{
+    public record EntityState
+    {
+        public string EntityId { get; init; } = "";
+
+        public string? State { get; init; }
```
helto4real, line 13: I wonder if we should have some convenience methods for translate to double and integer?

Model3/ModelV3/EntityStateCache.cs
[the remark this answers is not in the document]
helto4real: Yea that is a bug for sure. It should never make an exception it should return null imo.

Model3/ModelV3/EntityState.cs
FrankBakkerNl: Yes that is an open queation indeed. There are a few ootions but I cannot decide which is best. Lets discuss it and do it in a new PR

PR #450, opened 2021-08-31
Yaml config should not instantiate apps

Zorgino:
## Proposed change

Yaml config is no more responsible for instantiating apps.
Aligned with single responsibility principle.
Secrets input is now mockable.

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist


src/Daemon/NetDaemon.Daemon/Daemon/CodeManager.cs
```diff
@@ -29,62 +28,47 @@ public sealed class CodeManager : IInstanceDaemonApp
...
                 {
-                    using var fileReader = File.OpenText(file);
-                    var yamlAppConfig = new YamlAppConfig(_loadedDaemonApps, fileReader, _yamlConfig, file, appInstantiator);
+                    var appId = appType.GetCustomAttribute<NetDaemonAppAttribute>()?.Id ?? appType.Name;
 
-                    foreach (var appInstance in yamlAppConfig.GetInstances())
-                    {
-                        result.Add(appInstance);
-                    }
+                    yield return appInstantiator.Instantiate(appType, appId);
```
Zorgino, line 60: not sure whether we need yield here.
it just makes code look cleaner

src/Daemon/NetDaemon.Daemon/Daemon/AppInstantiator.cs
```diff
@@ -1,20 +1,35 @@
...
-            return new ApplicationContext(applicationType, appId, ServiceProvider);
+            try
+            {
+                return new ApplicationContext(applicationType, appId, ServiceProvider);
+            }
+            catch (Exception e)
+            {
+                var message = $"Error instantiating app of type {applicationType} with id \"{appId}\"";
+
+                _logger.LogTrace(e, $"{message}, use trace flag for details");
```
FrankBakkerNl, line 29: Why both trace and error? Ususlly I use just error and pass the exceotion iso just the message

src/Daemon/NetDaemon.Daemon/Daemon/CodeManager.cs
```diff
@@ -29,62 +28,48 @@ public sealed class CodeManager : IInstanceDaemonApp
...
             _yamlConfig = yamlConfig;
         }
 
         [SuppressMessage("", "CA1065")]
-        public int Count => _loadedDaemonApps.Count();
+        public int Count => DaemonAppTypes.Count();
 
         // Internal for testing
-        public IEnumerable<Type> DaemonAppTypes => _loadedDaemonApps;
+        internal IEnumerable<Type> DaemonAppTypes { get; }
```
FrankBakkerNl, line 39: Maybe the entire class can be internal (as making just this method internal is breaking anyway)

src/Daemon/NetDaemon.Daemon/Daemon/Config/IIoWrapper.cs
```diff
@@ -0,0 +1,7 @@
+namespace NetDaemon.Daemon.Config
+{
+    public interface IIoWrapper
```
FrankBakkerNl, line 3: Internal. In general I would like to only make things visible exteral when we really need it and make sure we want to commit to the way the API looks like

src/Daemon/NetDaemon.Daemon/Daemon/CodeManager.cs
```diff
@@ -29,62 +28,48 @@ public sealed class CodeManager : IInstanceDaemonApp
...
             {
                 _logger.LogWarning("No yaml configuration files or loaded apps found");
-                return result;
+                yield break;
             }
 
-            var appInstantiator = new AppInstantiator(serviceProvider);
-            
-            foreach (string file in allConfigFilePaths)
+            var appInstantiator = new AppInstantiator(serviceProvider, _logger);
```
FrankBakkerNl, line 50: I am not sure I follow what is going on here. What I suggested on Discord was:

If a yaml config references a class It should load that class without it needing the [NetDaemon} attribute
If a class has the [NetDaemonApp] apptribute it should be loaded as a single app without config

src/Daemon/NetDaemon.Daemon/Daemon/CodeManager.cs
FrankBakkerNl: I ususally use Linq in these sorts of cases. because I do not like a loop adding to a List. In this case it would probably be a Select(). where().ToList() If the lambda's het to complicated I use private method for them with specific descriptive names

src/Daemon/NetDaemon.Daemon/Daemon/AppInstantiator.cs
Zorgino: behavior from previous revision of CodeManager.cs
          
`
 _logger.LogTrace(e, "Error instance the app from the file {file}", file);

 _logger.LogError("Error instance the app from the file {file}, use trace flag for details", file);

  throw new NetDaemonException($"Error instance the app from the file {file}", e);
`

Zorgino: I am also interested why is it so, but decided not to change it not to break logging

src/Daemon/NetDaemon.Daemon/Daemon/CodeManager.cs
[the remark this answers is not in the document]
Zorgino: it's legacy
image

src/Daemon/NetDaemon.Daemon/Daemon/Config/IIoWrapper.cs
Zorgino: > make sure we want to commit to the way the API looks like
could you please elaborate?

src/Daemon/NetDaemon.Daemon/Daemon/CodeManager.cs
Zorgino: sorry, forgot to add "continue".

> If a yaml config references a class It should load that class without it needing the [NetDaemon} attribute
If a class has the [NetDaemonApp] apptribute it should be loaded as a single app without config

will add a test that will check that

Zorgino: it does exactly that

src/Daemon/NetDaemon.Daemon/Daemon/AppInstantiator.cs
FrankBakkerNl: Ah I see, but that was slightly different
_logger.LogTrace(e, "Error instance the app from the file {file}", file);
_logger.LogError("Error instance the app from the file {file}, use trace flag for details", file);
`
So the Error says you should turn on trace to get more deatils

FrankBakkerNl (approved):

PR #460, opened 2021-09-08
Add new codegen

Zorgino:
## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist


src/DaemonRunner/DaemonRunner/Service/App/CodeGeneration/NewCodeGenerator.Entitites.cs
```diff
@@ -0,0 +1,166 @@
...
+                        }
+
+                        attributes.Add(attributeName, TypeHelper.GetType(attributeObject));
+                    }
+                }
+
+                IEnumerable<(string Name, string TypeName, string SerializationName)> autoPropertiesParams = attributes
+                    .Select(a => (a.Key.ToNormalizedPascalCase(), a.Value.GetFriendlyName(), a.Key));
+
+                // handles the case when attributes have equal names in PascalCase but different types.
```
FrankBakkerNl, line 95: This might cause a problem when first there is only one such field and the user generates code, and later anther entity adds another with a differt type, the name of the first changes and breaks his exisisting automations

src/DaemonRunner/DaemonRunner/Service/App/CodeGeneration/NewCodeGenerator.ExtensionMethods.cs
```diff
@@ -0,0 +1,71 @@
...
+            {
+                var domain = domainServicesGroup.Key!;
+                var domainServices = domainServicesGroup
+                    .SelectMany(services => services!)
+                    .Where(s => s.Target?.Entity?.Domain != null)
+                    .Select(group => group)
+                    .OrderBy(x => x.Service)
+                    .ToList();
+
+                yield return GenerateDomainExtensionClass(domain, domainServices);
```
FrankBakkerNl, line 32: Suggestion: Instead of foreach and yield I would use a Select

src/DevelopmentApps/apps/CallServiceApp.cs
```diff
@@ -7,6 +7,7 @@
 namespace NetDaemon.DevelopmentApps.apps
 {
     [NetDaemonApp]
+    [Focus]
```
FrankBakkerNl, line 10: I keep forgetting myself to remove Focus from the PR

src/DevelopmentApps/apps/InterfaceApp/InterfaceApp.cs
```diff
@@ -11,19 +11,16 @@ namespace NetDaemon.DevelopmentApps.apps.DebugApp
     public class InterfaceApp : IAsyncInitializable
     {
         private INetDaemon _host;
-        private readonly ILogger _logger;
 
-        public InterfaceApp(INetDaemonHost host, ILogger logger)
+        public InterfaceApp(INetDaemonHost host)
```
FrankBakkerNl, line 15: Yep this no longer works

src/Model3/NetDeamon.Model3/Common/IHaContext.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using System;
+using JoySoftware.HomeAssistant.Model;
```
FrankBakkerNl, line 2: Add a new HassTarget class in the Model3 project so the app does not need te reference HassClient

src/DaemonRunner/DaemonRunner/Service/App/CodeGeneration/Extensions/TypeExtensions.cs
```diff
@@ -0,0 +1,61 @@
+﻿using System;
+using System.Collections.Generic;
+namespace NetDaemon.Service.App.CodeGeneration.Extensions
+{
+    internal static class TypeExtensions
+    {
+        private static readonly Dictionary<Type, string> _typeToFriendlyName = new()
+        {
+            { typeof(int), "int" },
```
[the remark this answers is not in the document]
Zorgino: this is for handling List<> and Dict<> which convert into List`1 and similar dict.
that cannot be parsed into type

src/DaemonRunner/DaemonRunner/Service/App/CodeGeneration/NewCodeGenerator.Entitites.cs
Zorgino: you're right.
what's the alternative of handling duplicates?
bear in mind, if we have attributes "Available" and "available" we cannot merge them, as each must have case sensitive jsonpropertyname.

src/DaemonRunner/DaemonRunner/Service/App/CodeGeneration/NewCodeGenerator.ExtensionMethods.cs
Zorgino: it would transform into long non readable LINQ (not LINQ to Entities) with `let`.

src/DevelopmentApps/apps/CallServiceApp.cs
Zorgino: that's from you commit I've cherry-picked  ;)
unintentionally pushed here

src/DaemonRunner/DaemonRunner/Service/App/CodeGeneration/NewCodeGenerator.Entitites.cs
Zorgino: one option would be to leave attributes in ha manner.
so we will have two attributes in SensorAttributes : lower and upper case props.
but if we have two same name attributes, but of different types. 
we can generate a single property for it of type object.
but if one is gone, during regeneration the other will get type by it's value.
and it breaks user's existing automations

src/DaemonRunner/DaemonRunner/Service/App/CodeGeneration/Extensions/TypeExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+            { typeof(long?), "long?" },
+            { typeof(ulong?), "ulong?" },
+            { typeof(float?), "float?" },
+            { typeof(double?), "double?" },
+            { typeof(bool?), "bool?" },
+            { typeof(void), "void" }
+        };
+
+        public static string GetFriendlyName(this Type type)
+        {
```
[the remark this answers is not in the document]
Zorgino: this is for handling List<> and Dict<> which convert into List`1 and similar dict.
that won't parse

src/DaemonRunner/DaemonRunner/Service/App/CodeGeneration/Extensions/CollectionExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+    internal static class CollectionExtensions
+    {
+        public static IEnumerable<TSource> Duplicates<TSource, TKey>(this IEnumerable<TSource> source, Func<TSource, TKey> selector)
+        {
+            var grouped = source.GroupBy(selector);
+            var moreThan1 = grouped.Where(i => i.IsMultiple());
+            return moreThan1.SelectMany(i => i);
+        }
+
+        public static IEnumerable<TSource> HandleDuplicates<TSource, TKey>(this IEnumerable<TSource> source, Func<TSource, TKey> selector, Func<TSource, TSource> func)
```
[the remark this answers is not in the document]
Zorgino: yes

src/DaemonRunner/DaemonRunner/Service/App/CodeGeneration/Helpers/HassServiceArgumentMapper.cs
```diff
@@ -0,0 +1,40 @@
...
+                Type = type,
+                Required = field.Required == true
+            };
+        }
+        private static Type GetTypeFromSelector(object? selectorObject)
+        {
+            return selectorObject switch
+            {
+                ActionSelector
+                    or AreaSelector
```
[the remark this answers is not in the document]
Zorgino: those are types

src/DaemonRunner/DaemonRunner/Service/App/CodeGeneration/NewCodeGenerator.Entitites.cs
```diff
@@ -0,0 +1,166 @@
...
+        {
+            foreach (var entityDomainGroups in entities.GroupBy(x => EntityIdHelper.GetDomain(x.EntityId)))
+            {
+                var attributes = new Dictionary<string, Type>();
+
+                foreach (var entity in entityDomainGroups)
+                {
+                    foreach (var (attributeName, attributeObject) in new Dictionary<string, object>(entity.Attribute))
+                    {
+                        if (attributes.ContainsKey(attributeName))
```
FrankBakkerNl, line 83: It looks like this is discarding duplicate names

src/DaemonRunner/DaemonRunner/Service/App/CodeGeneration/Helpers/HassServiceArgumentMapper.cs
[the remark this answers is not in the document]
FrankBakkerNl: SInce the default is now string this could be simplified by removing all that are explicitly mapped to string

src/DaemonRunner/DaemonRunner/Service/App/CodeGeneration/Helpers/HassServiceArgumentMapper.cs
```diff
@@ -0,0 +1,40 @@
...
+                    or EntitySelector
+                    or DeviceSelector
+                    or ObjectSelector
+                    or TargetSelector
+                    or TextSelector
+                    or null => typeof(string),
+                BooleanSelector => typeof(bool),
+                NumberSelector => typeof(long),
+                TimeSelector => typeof(DateTime),
+                SelectSelector => typeof(List<string>),
```
FrankBakkerNl, line 35: I found out this should be a string iso List<String> because it is a dropdown list that results in a single value

helto4real (changes requested):
Great work on this. Think @FrankBakkerNl caught alot of the details. Think tests should be added. Even if we do not add unit tests for every bit of functionality the code generation could be tested against real code provided in fake test files and fake data to generate them.

PR #500, opened 2021-11-02
Scheduling extension support

helto4real:
## Proposed change

Adds back scheduling support in the new model.

- [x] Add scheduling
- [x] Fix Dispose call bug
- [x] Add tests
- [x] Add push to nuget

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist


src/Daemon/NetDaemon.Daemon/DependencyInjectionSetup.cs
```diff
@@ -29,7 +30,7 @@ public static IServiceCollection AddNetDaemonServices(this IServiceCollection se
             // these interfaces and methods directly on INetDaemonHost can eventually be removed
             services.AddSingleton(s => s.GetRequiredService<NetDaemonHost>().HassEventsObservable);
             services.AddSingleton<ITextToSpeechService>(s => s.GetRequiredService<NetDaemonHost>().TextToSpeechService);
-            
+            services.AddTransient<INetDaemonScheduler, NetDaemonScheduler>();
```
BeeHiveJava, line 33: We shouldn't do this here. We need to have a `AddNetDaemonScheduler` method inside the scheduler extension. This way the Scheduler extension decides what and how dependencies get added, and whatever consuming project  (in this case NetDaemon) just needs to call that method.

helto4real: Good point

src/Extensions/NetDaemon.Extensions.Scheduling/NetDaemonScheduler.cs
```diff
@@ -0,0 +1,240 @@
...
+                .AddConsole();
+        });
+
+        private ILogger _log { get; }
+
+        /// <summary>
+        ///     Constructor
+        /// </summary>
+        /// <param name="loggerFactory">Injected logger factory</param>
+        public NetDaemonScheduler(ILoggerFactory? loggerFactory = null)
```
BeeHiveJava, line 32: Why don't we just inject an `ILogger<NetDaemonScheduler>` here?

src/App/NetDaemon.App/Common/ApplicationContexts/ApplicationContext.cs
```diff
@@ -157,6 +157,11 @@ public async ValueTask DisposeAsync()
                 await asyncDisposable.DisposeAsync().ConfigureAwait(false);
             }
 
+            if (ApplicationInstance is IDisposable disposable)
```
FrankBakkerNl, line 160: Yes this one was missing indeed

src/DevelopmentApps/apps/HassModel/Scheduling/ScheduleApp.cs
```diff
@@ -0,0 +1,27 @@
...
+using NetDaemon.DevelopmentApps.Domains.Sensor;
+using NetDaemon.DevelopmentApps.Domains.Zone;
+using NetDaemon.Extensions.Scheduler;
+using NetDaemon.HassModel.Common;
+using NetDaemon.HassModel.Entities;
+
+namespace NetDaemon.DevelopmentApps.apps.M3Test
+{
+    [NetDaemonApp]
+    public class ScheduledApp //: IDisposable, IAsyncDisposable
```
FrankBakkerNl, line 16: The comented code can now be removed

src/Extensions/NetDaemon.Extensions.Scheduling.Tests/LoggerMock.cs
```diff
@@ -0,0 +1,90 @@
+using Microsoft.Extensions.Logging;
+using Moq;
+using System;
+
+namespace NetDaemon.Extensions.Scheduling.Tests
+{
+    /// <summary>
+    ///     Logger mock
+    /// </summary>
+    public class LoggerMock
```
FrankBakkerNl, line 10: internal? If you want to use this in other test projects as well it would not be logical to put it in this project

src/Extensions/NetDaemon.Extensions.Scheduling/NetDaemonScheduler.cs
helto4real: I did not know the injection are that smart. I need to read up more on this  stuff.

src/Extensions/NetDaemon.Extensions.Scheduling.Tests/LoggerMock.cs
helto4real: Yea I was thinking about this. Maybe we could have a common test utils project.

src/Extensions/NetDaemon.Extensions.Scheduling/DisposableTimer.cs
```diff
@@ -0,0 +1,37 @@
...
+        private readonly CancellationTokenSource _combinedToken;
+        private readonly CancellationTokenSource _internalToken;
+        /// <summary>
+        ///     Constructor
+        /// </summary>
+        /// <param name="token">App cancellation token to combine</param>
+        public DisposableTimer(CancellationToken token)
+        {
+            _internalToken = new CancellationTokenSource();
+            _combinedToken = CancellationTokenSource.CreateLinkedTokenSource(_internalToken.Token, token);
```
FrankBakkerNl, line 20: I do not get the purpose of the Combined token here, cant the Dispose method here just dispose the token that is passed?

src/Extensions/NetDaemon.Extensions.Scheduling/INetDaemonScheduler.cs
```diff
@@ -0,0 +1,50 @@
...
+        ///     Cancel all scheduled timers on this instance
+        /// </summary>
+        void CancelAllTimers();
+
+        /// <summary>
+        ///     Run daily at a specific time
+        /// </summary>
+        /// <param name="time">The time in "hh:mm:ss" format</param>
+        /// <param name="action">Action to run</param>
+        IDisposable RunDaily(string time, Action action);
```
FrankBakkerNl, line 20: Passing a string for a time seems odd to me. A user needs to check what format is exepcted / supported etc.

I would actually not have all these overloads on the interface, just one that takes a timespan  or maybe some combination of interval / starttime kind of like how the reactive IScheduler works.

https://docs.microsoft.com/en-us/previous-versions/dotnet/reactive-extensions/hh229149(v=vs.103)

It has just a  few basic operations and then adds extension methods on top of that. The benefit is that if you want to implement the interface in anther class you only need to implement the core methods and get the rest for free

src/Extensions/NetDaemon.Extensions.Scheduling/NetDaemonScheduler.cs
```diff
@@ -0,0 +1,243 @@
...
+using System.Reactive.Linq;
+using System.Threading;
+using Microsoft.Extensions.Logging;
+
+namespace NetDaemon.Extensions.Scheduler
+{
+    /// <summary>
+    ///     Provides scheduling capability to be injected into apps
+    /// </summary>
+    public class NetDaemonScheduler : INetDaemonScheduler, IDisposable
```
FrankBakkerNl, line 14: If you set up the DI in this project as @BeeHiveJava suggested, this class could be internal.

helto4real: Yes in deed

src/Extensions/NetDaemon.Extensions.Scheduling/NetDaemon.Extensions.Scheduling.csproj
```diff
@@ -0,0 +1,37 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
FrankBakkerNl, line 1: I did not expect you made a separate project just for the schedulet, more like a generic Extensions project that contains the scheduler and maybe more in the future. Not sure what I think is better.  Making new projects for each extension feels like a lot of extra work to maintain scripts and nugets etc.

src/Extensions/NetDaemon.Extensions.Scheduling/DisposableTimer.cs
helto4real: The token passed is a token that cancels all times when the scheduler disposes. The internal one if to stop the current running timer

src/Extensions/NetDaemon.Extensions.Scheduling/NetDaemon.Extensions.Scheduling.csproj
helto4real: It is not that much maintenance extra costs imo. It is a few lines at most. I am afraid we will get cluttered packages again.

src/Extensions/NetDaemon.Extensions.Scheduling/NetDaemonScheduler.cs
```diff
@@ -0,0 +1,199 @@
...
+        [SuppressMessage("", "CA1031")]
+        public IDisposable RunAt(DateTimeOffset timeOffset, Action action)
+        {
+            var result = new DisposableTimer(_cancelTimers.Token);
+
+            Observable.Timer(
+                timeOffset,
+                _reactiveScheduler)
+                .Subscribe(
+                    _ =>
```
FrankBakkerNl, line 149: [Suggestion}
I think this is 4 times exacly the same, so maybe a private method would be nice here. If it is not exactly the same it would help to see more clearly what the difference is

src/Extensions/NetDaemon.Extensions.Scheduling/NetDaemonScheduler.cs
```diff
@@ -0,0 +1,199 @@
...
+                    result.Token
+                    );
+
+            return result;
+        }
+
+        /// <summary>
+        ///     Dispose the runnting timers correctly by cancel them
+        /// </summary>
+        protected virtual void Dispose(bool disposing)
```
FrankBakkerNl, line 175: [Suggestion] As an alternative to the 'full Dispose pattern' you can also make the class sealed. This meachanism is really only needed for classes that are intended to be inherited from. If the class is sealed the code analiyers will also sop comlaining if you do not use this pattern

FrankBakkerNl (approved):

src/Extensions/NetDaemon.Extensions.Scheduling/NetDaemonScheduler.cs
helto4real: You mean inside the subscribe? That is a good idea.

helto4real: fixed

src/Extensions/NetDaemon.Extensions.Scheduling/NetDaemonScheduler.cs
helto4real: Fixed

PR #542, opened 2021-12-09
Hass model areas

parkerman92:
## Proposed change

Fixes an issue in APIv2 where area is populated with the area id instead of the name (which I think is incorrect)
Adds area name to the new hass model api.

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/src/HassModel/NetDeamon.HassModel/Common/IHaContext.cs
+++ b/src/HassModel/NetDeamon.HassModel/Common/IHaContext.cs
@@ -1,7 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Reactive.Linq;
-using JoySoftware.HomeAssistant.Model;
 using NetDaemon.HassModel.Entities;
 
 namespace NetDaemon.HassModel.Common
@@ -46,6 +44,6 @@ namespace NetDaemon.HassModel.Common
         /// </summary>
         /// <param name="entityId"></param>
         /// <returns></returns>
-        HassArea? GetArea(string entityId);
+        Area? GetAreaFromEntityId(string entityId);
     }
 }
--- a/src/HassModel/NetDeamon.HassModel/Common/IHaContext.cs
+++ b/src/HassModel/NetDeamon.HassModel/Common/IHaContext.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Reactive.Linq;
+using JoySoftware.HomeAssistant.Model;
 using NetDaemon.HassModel.Entities;
 
 namespace NetDaemon.HassModel.Common
@@ -39,5 +40,12 @@ namespace NetDaemon.HassModel.Common
         /// <param name="target">The target that is targeted by this service call</param>
         /// <param name="data">Data provided to service. Should be Json-serializable to the data expected by the service</param>
         void CallService(string domain, string service, ServiceTarget? target = null, object? data = null);
+
+        /// <summary>
+        /// Get area for a single entity
+        /// </summary>
+        /// <param name="entityId"></param>
+        /// <returns></returns>
+        HassArea? GetArea(string entityId);
     }
 }
--- a/src/HassModel/NetDeamon.HassModel/Internal/EntityAreaCache.cs
+++ b/src/HassModel/NetDeamon.HassModel/Internal/EntityAreaCache.cs
@@ -1,0 +1,83 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using JoySoftware.HomeAssistant.Client;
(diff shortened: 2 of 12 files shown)


src/HassModel/NetDeamon.HassModel/Internal/EntityAreaCache.cs
```diff
@@ -0,0 +1,83 @@
...
+
+    private async Task LoadAreas()
+    {
+        var entities = await _hassClient.GetEntities().ConfigureAwait(false);
+        var devices = await _hassClient.GetDevices().ConfigureAwait(false);
+        var deviceDict = devices?.ToDictionary(k => k.Id!, v => v);
+        var areas = await _hassClient.GetAreas().ConfigureAwait(false);
+        var areaDict = areas?.ToDictionary(k => k.Id!, v => v);
+
+        _latestAreas.Clear();
```
FrankBakkerNl, line 52: Instead of Clearing the dictionary and re-loading you could create a whole new dictionary, load it and then re-assign the private field to the new dictionary when it is fully loaded. This will prevent any possible concurrency issue between clearing and re-filling it. You then don't even need a ConcurrentDictionary because it is only ever written to initially by one thread

src/HassModel/NetDeamon.HassModel/Internal/EntityAreaCache.cs
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    public HassArea? GetArea(string entityId)
+    {
+        if (!_initialized) throw new InvalidOperationException("AreaCache has not been initialized yet");
+
+        return _latestAreas.TryGetValue(entityId, out var result) ? result : null;
+    }
+
+    public void Dispose()
```
FrankBakkerNl, line 39: Personally I like to put Dispose() last

src/HassModel/NetDeamon.HassModel/Common/IHaContext.cs
```diff
@@ -39,5 +40,12 @@ public interface IHaContext
         /// <param name="target">The target that is targeted by this service call</param>
         /// <param name="data">Data provided to service. Should be Json-serializable to the data expected by the service</param>
         void CallService(string domain, string service, ServiceTarget? target = null, object? data = null);
+
+        /// <summary>
+        /// Get area for a single entity
+        /// </summary>
+        /// <param name="entityId"></param>
+        /// <returns></returns>
+        HassArea? GetArea(string entityId);
```
FrankBakkerNl, line 49: Is this the HassArea class from the HassClient? I prefer not to expose that in the HassModel public API. HassModel currenty only uses HassClient internally.

I was thinking of a design where IHaContext returns some more metadata about an Entity in the form of an `EntityMetaData` object that has Device, Area etc. That will however be a bit more complicated as we need to design such a structure which is basically why this Area was not implemented yet.

FrankBakkerNl:
I really like this It is almost exactly what I had in mind for this.
Just the public API of IHaContext I would not like to depend on HassClient objects Lets discuss on Discord on how exactly to make the API

src/HassModel/NetDeamon.HassModel/Internal/EntityAreaCache.cs
parkerman92: Added - [Commit](https://github.com/net-daemon/netdaemon/pull/542/commits/015031873feae0ba2e0b6a733f439b047f8b359b)

src/HassModel/NetDeamon.HassModel/Internal/EntityAreaCache.cs
parkerman92: Moved

src/HassModel/NetDeamon.HassModel/Common/IHaContext.cs
parkerman92: As discussed on discord. Refactored to use simple class and not expose the HassClient models.

parkerman92:
Cheers for the feedback @FrankBakkerNl any other suggestions let me know 👍

FrankBakkerNl (approved):

src/HassModel/NetDeamon.HassModel/Internal/HassObjectMapper.cs
```diff
@@ -58,5 +58,14 @@ public static Event Map(this HassEvent hassEvent)
                 DataElement = hassEvent.DataElement,
             };
         }
+
+        public static Area Map(this HassArea hassArea)
+        {
+            return new Area()
+            {
+                // Can an area not have a name?
+                Name = hassArea.Name!,
```
FrankBakkerNl, line 67:
```suggestion
                Name = hassArea.Name,
```

helto4real:
Thank you for the PR!

PR #556, opened 2021-12-19
Adds future new AppModel to the NetDaemon project

helto4real:
## Proposed change

This is the new AppModel in NetDaemon that will replace the NetDaemon one eventually. This model has a much clearer purpose to handle application lifecykle and configuration for NetDaemon apps if or not if they use HassModel.
This is what is added:

- [x] New application lifecycle management that has better design
- [x] New configuration model that uses standard c# dotnet configuration that are injected into apps
- [x] Support both dynamic compiled apps and statically compiled apps
- [x] Inject convertable types from string in config
- [ ] Comment all public interfaces and classes
- [ ] Add Hosts and docker images for v3 standard and add-on
- [ ] Support add services to apps in this model too

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist


src/AppModel/NetDaemon.AppModel.Tests/AppModelTests.cs
```diff
@@ -0,0 +1,139 @@
...
+
+        // CHECK
+        loadApps.Should().HaveCount(1);
+    }
+
+    [Fact]
+    public async Task TestGetApplicationsWithCompilerError()
+    {
+        // ARRANGE
+        var builder = Host.CreateDefaultBuilder()
```
FrankBakkerNl, line 45: This is quite repetitive, maybe create a (private) helper for this

helto4real: Aight good idea

src/AppModel/NetDaemon.AppModel/Internal/AppModelContext.cs
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    private async Task LoadApplications()
+    {
+        foreach (var appType in _applicationTypes)
+        {
+            var appAttribute = appType.GetCustomAttribute<NetDaemonAppAttribute>();
+            var id = appAttribute?.Id ?? appType.FullName ??
+                throw new InvalidOperationException("Type was not expected to be null");
+            var app = ActivatorUtilities.CreateInstance<Application>(_provider, id, appType);
```
FrankBakkerNl, line 50: Why do you need ActivatorUtilities here? I would think this can just be a `new Application()`

helto4real: Not sure. I had that I think first. I think it has ILogger<IApplication> that I did not want to create in advance. I saved a couple of lines using it

PR #594, opened 2022-01-11
Adds [Focus] attribute support

helto4real:
## Proposed change

Todo: tests

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist


src/AppModel/NetDaemon.AppModel/Internal/Extensions/TypeExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+internal static class TypeExtensions
+{
+    public static bool HasNetDaemonFocusAttribute(this Type type)
+    {
+        return Attribute.GetCustomAttribute(type, typeof(FocusAttribute)) is not null;
+    }
+
+    public static bool IsNetDaemonFocusAttributeUsed(this IEnumerable<Type> types)
+    {
+        return types.Any(n => n.IsClass &&
```
FrankBakkerNl, line 14: Are these checks al needed? If you only check for this attribute on NetDaemonApps I think you only need to check for the attribute. In which case these tow methods become pretty simple and are probably better suited as private methods  in the class that uses them than as Extension Methods

src/AppModel/NetDaemon.AppModel/Internal/AppModelContext.cs
```diff
@@ -25,29 +26,36 @@ internal class AppModelContext : IAppModelContext, IAsyncInitializable
...
     private async Task LoadApplications()
     {
+        var loadOnlyFocusedApps = _applicationTypes.IsNetDaemonFocusAttributeUsed();
         foreach (var appType in _applicationTypes)
         {
             var appAttribute = appType.GetCustomAttribute<NetDaemonAppAttribute>();
             var id = appAttribute?.Id ?? appType.FullName ??
                 throw new InvalidOperationException("Type was not expected to be null");
-            var app = ActivatorUtilities.CreateInstance<Application>(_provider, id, appType);
+            var loadMode = AppLoadMode.UseStateManager;
```
FrankBakkerNl, line 53: Instead of the LoadMode would not it be easier to just skip the NON-Focus mode apps if there is at least one Focus app present?

dev/DebugHost/apps/HelloApp/HelloApp.cs
```diff
@@ -7,6 +7,7 @@
 namespace Apps;
 
 [NetDaemonApp]
+[Focus]
```
FrankBakkerNl, line 10: remove

src/Runtime/NetDaemon.Runtime/Internal/AppStateManager.cs
```diff
@@ -21,83 +19,63 @@ IServiceProvider provider
...
-        using var scope = _provider.CreateScope();
-
-        var haContext = scope.ServiceProvider.GetRequiredService<IHaContext>();
-        var entityId = ToSafeHomeAssistantEntityIdFromApplicationId(applicationId);
+        var scope = _provider.CreateScope();
+        try
+        {
+            var haContext = scope.ServiceProvider.GetRequiredService<IHaContext>();
+            var entityId = ToSafeHomeAssistantEntityIdFromApplicationId(applicationId);
+            var appState = haContext.GetState(entityId);
```
FrankBakkerNl, line 31: Maybe it is easier to use the Client directly here instead of HaContext so you don't need to do the scope / dispose thing

helto4real: I wanted the cache features. GetState is not implemented in the new client since there is no use-case to get state from non-cache imo.

src/AppModel/NetDaemon.AppModel/Internal/Extensions/TypeExtensions.cs
helto4real: Well it is a filtering thing. Attributes cannot live on non classes/methods etc. I think it is valid.

FrankBakkerNl (approved):

PR #604, opened 2022-01-14
Added support for async initialization

helto4real:
## Proposed change

Adds the possibility for applications to init async operations.
- It is InitializeAsync is called on loading of the applications and will be awaited
- A warning if init takes too long
- A cancellationtoken will be provided to use when application is unloaded

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist


src/AppModel/NetDaemon.AppModel.Tests/Fixtures/Local/LocalAppWithInitializeAsync.cs
```diff
@@ -0,0 +1,14 @@
+namespace LocalApps;
+
+[NetDaemonApp]
+public class MyAppLocalAppWithInitializeAsync : IInitializableAsync
```
FrankBakkerNl, line 4: IAsyncInitializable is more consistent with IAsyncDisposable and with the current version

src/AppModel/NetDaemon.AppModel/Internal/Context/ApplicationContext.cs
```diff
@@ -2,6 +2,7 @@ namespace NetDaemon.AppModel.Internal;
 
 internal sealed class ApplicationContext
 {
+    private readonly CancellationTokenSource _cancelTokenSource = new();
```
FrankBakkerNl, line 5: This seems to never be fired, maybe better to pass a token into InitializeAsync()

helto4real: Yea I forgot to fire it on DisposeAsync. Good catch

src/AppModel/NetDaemon.AppModel.Tests/Fixtures/Local/LocalAppWithInitializeAsync.cs
helto4real: Lool correct!

src/AppModel/NetDaemon.AppModel/Internal/Context/ApplicationContext.cs
```diff
@@ -23,18 +24,29 @@ public ApplicationContext(Type appType, IServiceProvider serviceProvider)
...
+            _cancelTokenSource.Cancel();
+
         if (Instance is IAsyncDisposable asyncDisposable) await asyncDisposable.DisposeAsync().ConfigureAwait(false);
 
         else if (Instance is IDisposable disposable) disposable.Dispose();
 
         if (_serviceScope is IAsyncDisposable serviceScopeAsyncDisposable)
             await serviceScopeAsyncDisposable.DisposeAsync().ConfigureAwait(false);
+
+        _cancelTokenSource.Dispose();
```
FrankBakkerNl, line 50: Maybe do this first so the init is cancelled before we dispose

FrankBakkerNl (changes requested):

src/AppModel/NetDaemon.AppModel/Internal/Context/ApplicationContext.cs
helto4real: Yea that is rifht. The cancellation token could be used some other places too in long running processes.

helto4real: I all ready do Cancel before. I do not want to dispose it so code use disposed cancellation token. First I cancel, do asyncdispose call then dispose the  cancellation source.  So I think that should be right?

FrankBakkerNl (approved):

PR #644, opened 2022-01-29
feat: introduce app instance factory to allow extending app instantiation logic 

BeeHiveJava:
## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist


src/AppModel/NetDaemon.AppModel/Internal/Instance/AppInstanceHelper.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace NetDaemon.AppModel.Internal.Resolver;
+
+internal static class AppInstanceHelper
+{
+    public static string GetAppId(Type type)
+    {
+        var attribute = type.GetCustomAttribute<NetDaemonAppAttribute>();
+        if (attribute is null)
+        {
+            throw new InvalidOperationException($"{type} has no {nameof(NetDaemonAppAttribute)}");
```
FrankBakkerNl, line 12: If it does notbhave the attribite you can still fallback to the type name

src/AppModel/NetDaemon.AppModel/Internal/Instance/IAppInstance.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace NetDaemon.AppModel.Internal.Resolver;
+
+internal interface IAppInstance
```
FrankBakkerNl, line 3: The name is a bit strange. This interfave does not represents an Instance but it creates instaces. The IAppInstanceResolver I think also does not have a correct name, it resolves apps, not instances of apps.

src/AppModel/NetDaemon.AppModel/Common/Extensions/ServiceCollectionExtension.cs
```diff
@@ -95,13 +118,22 @@ internal static IServiceCollection AddAppModelIfNotExist(this IServiceCollection
 
     internal static IServiceCollection AddAppTypeResolverIfNotExist(this IServiceCollection services)
     {
-        // Check if we already registered
-        if (services.Any(n => n.ImplementationType == typeof(AppTypeResolver)))
-            return services;
+        // Check if already registered
+        if (services.All(descriptor => descriptor.ImplementationType != typeof(AppTypeResolver)))
```
FrankBakkerNl, line 122: Personally I find the Any(.... == ) cleare than the All(.. !=  )  in this case

BeeHiveJava: But it's different. I inversed the logic here.

BeeHiveJava: The comment is outdated.

FrankBakkerNl: Ah but then I would prefer
if(!sercvies.Any(...))
I know Resharper.Rider will suggests the All here but I think it is strange to think of it that way. We do not do this because all registerd types are not of a specific type, but there is not any there yet that is of this type

src/AppModel/NetDaemon.AppModel/Internal/AppAssemblyProvider/SingleAppAssemblyProvider.cs
```diff
@@ -0,0 +1,18 @@
+using System.Reflection;
+
+namespace NetDaemon.AppModel.Internal.AppAssemblyProvider;
+
+internal class SingleAppAssemblyProvider : IAppAssemblyProvider
```
FrankBakkerNl, line 5: This is not just fo single apps (they dont even Use this provider I think) Maybe this should just be called AppAssemblyProvider

BeeHiveJava: Then it will clash with the namespace. Maybe `DefaultAppAssemblyProvider`?

BeeHiveJava: Or change the namespaces, I don't really like them anyway.

BeeHiveJava: Renamed namespaces to be plural + renamed this class.

src/AppModel/NetDaemon.AppModel/Internal/AppFactories/AppFactoryHelper.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System.Reflection;
+
+namespace NetDaemon.AppModel.Internal.AppFactories;
+
+internal static class AppFactoryHelper
```
FrankBakkerNl, line 5: As there is just one place these are used now they could also be private memebers

BeeHiveJava: Used in `TypeAppFactory` and `FuncAppFactory`. One gets it based on a generic parameter, the other on a specifically passed `Type` parameter.

src/AppModel/NetDaemon.AppModel/Internal/AppFactories/TypeAppFactory.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace NetDaemon.AppModel.Internal.AppFactories;
+
+internal class TypeAppFactory : FuncAppFactory
+{
+    public TypeAppFactory(Type type, string? id = default, bool? focus = default) :
+        base(
+            CreateFactoryFunc(type),
+            id ?? AppFactoryHelper.GetAppId(type),
```
FrankBakkerNl, line 8: These are not needed hare as the base class also does that

BeeHiveJava: Only the generic base class does this. I can move it to the non-generic base class, but then I'll need to work reflection magic to somehow get the returned `Type` from the `Func`.

BeeHiveJava: Or pass the `Type` parameter in the constructor, but I don't like that because in theory it could differ from the `Type` in the `Func`.

BeeHiveJava: Basically I only use these functions when I know the type is correct. That is two scenario's:
- You use the `FuncAppFactory` generic variant. In that case we enforce the generic parameter in the `Func`.
- You use the `TypeAppFactory`, in this case you specify the `Type` explicitly.

The third scenario is that you use the non-generic `FuncAppFactory`, but in that case you'll have to just pass the id and focus, because there's no reliable way to get it.

src/AppModel/NetDaemon.AppModel.Tests/Fixtures/Local/LocalAppWithId.cs
```diff
@@ -1,6 +1,7 @@
 namespace LocalApps;
 
-[NetDaemonApp(Id = "SomeId")]
```
helto4real, line 3: [Why this change? Before we tested that user provided id was set byt the appmodel to the app. Maybe I am missing something :)
[Edit: Never mind you just use a const instead of hardcoded]

helto4real (approved):
I think you did an awesome job. The tests look really clean too!

src/AppModel/NetDaemon.AppModel/Common/IAppInstanceFactory.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace NetDaemon.AppModel;
+
+public interface IAppInstanceFactory
```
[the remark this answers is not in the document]
BeeHiveJava: No, this should be removed.

PR #653, opened 2022-02-01
Add background task tracking and proper error handling

helto4real:
## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/src/HassModel/NetDaemon.HassModel.Tests/Internal/AppScopedHaContextProviderTest.cs
+++ b/src/HassModel/NetDaemon.HassModel.Tests/Internal/AppScopedHaContextProviderTest.cs
@@ -11,6 +11,7 @@ using NetDaemon.Client;
 using NetDaemon.Client.HomeAssistant.Model;
 using NetDaemon.Client.Internal.HomeAssistant.Commands;
 using NetDaemon.HassModel.Entities;
+using NetDaemon.HassModel.Internal;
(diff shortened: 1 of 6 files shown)


src/HassModel/NetDaemon.HassModel.Tests/Internal/AppScopedHaContextProviderTest.cs
```diff
@@ -168,6 +169,34 @@ public async Task EventsStopAfterDispose()
         eventObserverMock.VerifyNoOtherCalls();
     }
 
+    [Fact]
+    public async Task TestThatCallServiceTrackBackgroundTask()
```
FrankBakkerNl, line 173: These tests really test the implementation, not really the desired result. Maybe better to mock the logger and check that its gets called on an exception after an await

FrankBakkerNl (approved):

src/HassModel/NetDaemon.HassModel.Tests/Internal/AppScopedHaContextProviderTest.cs
helto4real: It is actually a unit test in combination with the test of the Background task tests that actually tests exceptions being logged. So the combination should test the desired result. We do test the logger in that test. So here only need to test that the background task being used.

PR #656, opened 2022-02-03
Initial commit of MQTT entity management framework.

skotl:
Addable extension method `AddMqttExtensions()`
Rough entity creation through `EntityUpdater` (which is probably a poor name) that needs more strongly-typed.

**Note that I can't push this to a branch on the netdaemon core, which I think it would be better to be in - if you want to create a new branch ("mqtt_extensions" or similar?) then I can update the PR to point to that instead.**

That will also allow Eugene and I to collaborate around the branch without upsetting your work.

This has been tested manually within the context of the `Create entities from ND` discord chat.

## Breaking change
Non-breaking change

## Proposed change
New extension project called `NetDaemon.Extensions.MqttEntities` that adds a services extension `AddMqttExtensions()` that can be called from program.cs

Sample usage:

```csharp
    public MqttTester(ILogger<MqttTester> logger, IEntityUpdater entityUpdater)
    {
        _logger = logger;
        _entityUpdater = entityUpdater;

        _entityUpdater.CreateAsync("binary_sensor", "motion", "motion_test_1", "mqtt test")
            .GetAwaiter();
    }
```

Requires additional config in `appSettings.json`:

```json
  "Mqtt": {
    "Host": "homeassistant.local",
    "Port": "1883",
    "UserId": "mqtt",
    "Password": "secret"
  }

```

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/NetDaemon.Extensions.MqttEntities/DependencyInjectionSetup.cs
+++ b/NetDaemon.Extensions.MqttEntities/DependencyInjectionSetup.cs
@@ -12,8 +12,8 @@ public static class DependencyInjectionSetup
     public static IServiceCollection AddMqttExtensions(this IServiceCollection services)
     {
         services.AddSingleton<IMqttFactory, MqttFactory>();
-        services.AddScoped<IEntityUpdater, EntityUpdater>();
-        services.AddScoped<IMessageSender, MessageSender>();
+        services.AddSingleton<IEntityUpdater, EntityUpdater>();
+        services.AddSingleton<IMessageSender, MessageSender>();
         
         return services;
     }
--- a/NetDaemon.Extensions.MqttEntities/DependencyInjectionSetup.cs
+++ b/NetDaemon.Extensions.MqttEntities/DependencyInjectionSetup.cs
@@ -1,0 +1,20 @@
+﻿using Microsoft.Extensions.DependencyInjection;
+using MQTTnet;
+
+namespace NetDaemon.Extensions.MqttEntities;
+
+public static class DependencyInjectionSetup
+{
+    /// <summary>
+    ///     Adds scheduling capabilities through dependency injection
+    /// </summary>
+    /// <param name="services">Provided service collection</param>
+    public static IServiceCollection AddMqttExtensions(this IServiceCollection services)
+    {
+        services.AddSingleton<IMqttFactory, MqttFactory>();
+        services.AddScoped<IEntityUpdater, EntityUpdater>();
+        services.AddScoped<IMessageSender, MessageSender>();
+        
+        return services;
+    }
+}
--- a/NetDaemon.Extensions.MqttEntities/IEntityUpdater.cs
+++ b/NetDaemon.Extensions.MqttEntities/IEntityUpdater.cs
@@ -1,0 +1,6 @@
+﻿namespace NetDaemon.Extensions.MqttEntities;
+
+public interface IEntityUpdater
+{
+    Task CreateAsync(string deviceType, string deviceClass, string entityId, string name);
+}
--- a/NetDaemon.Extensions.MqttEntities/IMessageSender.cs
+++ b/NetDaemon.Extensions.MqttEntities/IMessageSender.cs
@@ -1,0 +1,6 @@
+﻿namespace NetDaemon.Extensions.MqttEntities;
+
+public interface IMessageSender
+{
+    Task SendMessageAsync(string topic, string payload);
+}
--- a/NetDaemon.Extensions.MqttEntities/MessageSender.cs
+++ b/NetDaemon.Extensions.MqttEntities/MessageSender.cs
@@ -1,0 +1,64 @@
+﻿using System.Security.Authentication;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+using MQTTnet;
+using MQTTnet.Client;
+using MQTTnet.Client.Connecting;
+using MQTTnet.Client.Options;
+using MQTTnet.Client.Publishing;
+
+namespace NetDaemon.Extensions.MqttEntities;
+
+internal class MessageSender : IMessageSender
+{
+    private readonly ILogger<MessageSender> _logger;
+    private readonly IMqttFactory _mqttFactory;
+    private readonly MqttConfiguration? _mqttConfig;
+    
+    public MessageSender(ILogger<MessageSender> logger, IConfiguration configuration, IMqttFactory mqttFactory)
+    {
+        _logger = logger;
+        _mqttFactory = mqttFactory;
+        
+        _mqttConfig = configuration.GetSection("Mqtt")
+            .Get<MqttConfiguration>();
+        
+        _logger.LogDebug($"MQTT connection is {_mqttConfig?.Host}:{_mqttConfig?.Port}/{_mqttConfig?.UserId}");
+    }
+
+    public async Task SendMessageAsync(string topic, string payload)
+    {
+        using (var mqttClient = _mqttFactory.CreateMqttClient())
+        {
+            await ConnectAsync(mqttClient);
+            await PublishMessage(mqttClient, topic, payload);
+        }
+    }
+
+    private async Task ConnectAsync(IMqttClient mqttClient)
+    {
+        var options = new MqttClientOptionsBuilder()
+            .WithTcpServer(_mqttConfig?.Host, _mqttConfig?.Port)
+            .WithCredentials(_mqttConfig?.UserId, _mqttConfig?.Password)
+            .Build();
+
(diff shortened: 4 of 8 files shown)


NetDaemon.Extensions.MqttEntities/DependencyInjectionSetup.cs
```diff
@@ -0,0 +1,20 @@
...
+public static class DependencyInjectionSetup
+{
+    /// <summary>
+    ///     Adds scheduling capabilities through dependency injection
+    /// </summary>
+    /// <param name="services">Provided service collection</param>
+    public static IServiceCollection AddMqttExtensions(this IServiceCollection services)
+    {
+        services.AddSingleton<IMqttFactory, MqttFactory>();
+        services.AddScoped<IEntityUpdater, EntityUpdater>();
```
FrankBakkerNl, line 15: Scoped is only usefull if you have state in tje class that should be scoped (per app in case of ND) I think thats not the case here.

helto4real:
@skotl I will add you to the dev team here. You can push to branch here your self and make it easier for you guys to collaborate

NetDaemon.Extensions.MqttEntities/DependencyInjectionSetup.cs
skotl: My concern is the connection to mqtt, which times out after 10 seconds.
But actually... I remake the connection each time when a message is called, so that's probably not an issue.

OK - will update - thanks

NetDaemon.Extensions.MqttEntities/DependencyInjectionSetup.cs
```diff
@@ -0,0 +1,20 @@
...
+
+public static class DependencyInjectionSetup
+{
+    /// <summary>
+    ///     Adds scheduling capabilities through dependency injection
+    /// </summary>
+    /// <param name="services">Provided service collection</param>
+    public static IServiceCollection AddMqttExtensions(this IServiceCollection services)
+    {
+        services.AddSingleton<IMqttFactory, MqttFactory>();
```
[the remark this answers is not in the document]
skotl: It's in the MQTTnet nuget package

skotl:
Changed DI scope to Singleton for the local services, changed target branch to mqtt_extensions.
Just need to check that @FrankBakkerNl  is happy with the comment on MqttFactory

NetDaemon.Extensions.MqttEntities/IEntityUpdater.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace NetDaemon.Extensions.MqttEntities;
+
+public interface IEntityUpdater
+{
+    Task CreateAsync(string deviceType, string deviceClass, string entityId, string name);
```
BeeHiveJava, line 5: Maybe it's worth it to create a value object here to avoid passing around many parameters.

BeeHiveJava: Also cancellation token support?

NetDaemon.Extensions.MqttEntities/IMessageSender.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace NetDaemon.Extensions.MqttEntities;
+
+public interface IMessageSender
+{
+    Task SendMessageAsync(string topic, string payload);
```
BeeHiveJava, line 5: Cancellation token support?

NetDaemon.Extensions.MqttEntities/MessageSender.cs
```diff
@@ -0,0 +1,64 @@
...
+    private readonly ILogger<MessageSender> _logger;
+    private readonly IMqttFactory _mqttFactory;
+    private readonly MqttConfiguration? _mqttConfig;
+    
+    public MessageSender(ILogger<MessageSender> logger, IConfiguration configuration, IMqttFactory mqttFactory)
+    {
+        _logger = logger;
+        _mqttFactory = mqttFactory;
+        
+        _mqttConfig = configuration.GetSection("Mqtt")
```
BeeHiveJava, line 23: Should use the options framework instead here.

NetDaemon.Extensions.MqttEntities/MessageSender.cs
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    private async Task ConnectAsync(IMqttClient mqttClient)
+    {
+        var options = new MqttClientOptionsBuilder()
+            .WithTcpServer(_mqttConfig?.Host, _mqttConfig?.Port)
+            .WithCredentials(_mqttConfig?.UserId, _mqttConfig?.Password)
+            .Build();
+
+        var connectResult = await mqttClient.ConnectAsync(options, CancellationToken.None);
```
BeeHiveJava, line 45: Why no cancellation token?

NetDaemon.Extensions.MqttEntities/MqttConfiguration.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace NetDaemon.Extensions.MqttEntities;
+
+internal class MqttConfiguration
+{
+    public string? Host { get; set; }
```
BeeHiveJava, line 5: Are these properties really nullable?

NetDaemon.Extensions.MqttEntities/IEntityUpdater.cs
skotl: My ambition was to get the extension framework in so that Eugene could take on the actual implentation (which is likely to be of the `<T> : where T is Entity` approach) so I wouldn't take this implementation of `IEntityUpdater` as representing how it will end up.
It's really just intended as an example of how to call the MQTT message sender.

NetDaemon.Extensions.MqttEntities/IMessageSender.cs
skotl: Do you think it's needed when the connection, message-send and disposal is all wrapped in a single atomic call?

NetDaemon.Extensions.MqttEntities/MqttConfiguration.cs
skotl: Probably not - I'll update - thanks :)

FrankBakkerNl: You can probably use a record here with a primary constructor and have the properties auto generated. That way you can make then non nullabe without getting warnings. Username and password might actually be optional

NetDaemon.Extensions.MqttEntities/MessageSender.cs
```diff
@@ -0,0 +1,64 @@
...
+        
+        _mqttConfig = configuration.GetSection("Mqtt")
+            .Get<MqttConfiguration>();
+        
+        _logger.LogDebug($"MQTT connection is {_mqttConfig?.Host}:{_mqttConfig?.Port}/{_mqttConfig?.UserId}");
+    }
+
+    public async Task SendMessageAsync(string topic, string payload)
+    {
+        using (var mqttClient = _mqttFactory.CreateMqttClient())
```
FrankBakkerNl, line 31: Maybe cache this client in a field so it is not recreated each time and then dispose in Dispose if this class.

NetDaemon.Extensions.MqttEntities/DependencyInjectionSetup.cs
```diff
@@ -0,0 +1,20 @@
...
+        services.AddSingleton<IEntityUpdater, EntityUpdater>();
+        services.AddSingleton<IMessageSender, MessageSender>();
```
FrankBakkerNl, line 16: The MessageSender does not have any state so it does not have to be a singleton. I read its recimended to use transient in those cases (creating a new instance is relatively cheap)

helto4real (approved):
I am ok for this to be merged to the feature branch so @eugeneniemand and @skotl can work on it. I save my comments to the real PR. You can use these comments as guidance for your work. Thanks for the contribution!

NetDaemon.Extensions.MqttEntities/IEntityUpdater.cs
BeeHiveJava: Fair enough!

NetDaemon.Extensions.MqttEntities/IMessageSender.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace NetDaemon.Extensions.MqttEntities;
+
+public interface IMessageSender
```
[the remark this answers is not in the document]
skotl: Yes - definitely should be internal!

PR #657, opened 2022-02-03
Automatically remove unused app state input booleans

helto4real:
## Proposed change

Removes unused input boolean helpers for non existing apps

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/src/Runtime/NetDaemon.Runtime/Internal/AppStateManager.cs
+++ b/src/Runtime/NetDaemon.Runtime/Internal/AppStateManager.cs
@@ -54,7 +54,7 @@ internal class AppStateManager : IAppStateManager, IHandleHomeAssistantAppStateU
                 }
             }).Subscribe();
     }
-    
+
     public async Task<ApplicationState> GetStateAsync(string applicationId)
     {
         if (_stateCache.TryGetValue(applicationId, out var applicationState)) return applicationState;
@@ -72,13 +72,14 @@ internal class AppStateManager : IAppStateManager, IHandleHomeAssistantAppStateU
         var isEnabled = await _appStateRepository.GetOrCreateAsync(applicationId, _cancelTokenSource.Token)
             .ConfigureAwait(false);
 
-        switch (state)
+        // Only update state if it is different from current
+        if (
+            (state == ApplicationState.Enabled && !isEnabled) ||
+            (state == ApplicationState.Disabled && isEnabled)
+            )
         {
-            case ApplicationState.Enabled when !isEnabled:
-            case ApplicationState.Disabled when isEnabled:
-                await _appStateRepository.UpdateAsync(applicationId, isEnabled, _cancelTokenSource.Token)
-                    .ConfigureAwait(false);
-                break;
+            await _appStateRepository.UpdateAsync(applicationId, isEnabled, _cancelTokenSource.Token)
+                .ConfigureAwait(false);
         }
     }
 
--- a/src/Runtime/NetDaemon.Runtime/Internal/AppStateManager.cs
+++ b/src/Runtime/NetDaemon.Runtime/Internal/AppStateManager.cs
@@ -17,38 +17,6 @@ internal class AppStateManager : IAppStateManager, IHandleHomeAssistantAppStateU
         _appStateRepository = appStateRepository;
     }
 
-    public async Task<ApplicationState> GetStateAsync(string applicationId)
-    {
-        if (_stateCache.TryGetValue(applicationId, out var applicationState)) return applicationState;
-
-        return await _appStateRepository.GetOrCreateAsync(applicationId, _cancelTokenSource.Token)
-            .ConfigureAwait(false)
-            ? ApplicationState.Enabled
-            : ApplicationState.Disabled;
-    }
-
-    public async Task SaveStateAsync(string applicationId, ApplicationState state)
-    {
-        _stateCache[applicationId] = state;
-
-        var isEnabled = await _appStateRepository.GetOrCreateAsync(applicationId, _cancelTokenSource.Token)
-            .ConfigureAwait(false);
-
-        switch (state)
-        {
-            case ApplicationState.Enabled when !isEnabled:
-            case ApplicationState.Disabled when isEnabled:
-                await _appStateRepository.UpdateAsync(applicationId, isEnabled, _cancelTokenSource.Token)
-                    .ConfigureAwait(false);
-                break;
-        }
-    }
-
-    public void Dispose()
-    {
-        _cancelTokenSource.Dispose();
-    }
-
     public async Task InitializeAsync(IHomeAssistantConnection haConnection, IAppModelContext appContext)
     {
         _stateCache.Clear();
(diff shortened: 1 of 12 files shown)


src/Runtime/NetDaemon.Runtime/Internal/AppStateManager.cs
```diff
@@ -1,67 +1,61 @@
...
-                await haConnection.CallServiceAsync("input_boolean", "turn_on",
-                    new HassTarget {EntityIds = new[] {entityId}},
-                    cancelToken: _cancelTokenSource.Token);
-                break;
-            case ApplicationState.Disabled when currentState == "on":
-                await haConnection.CallServiceAsync("input_boolean", "turn_off",
-                    new HassTarget {EntityIds = new[] {entityId}},
-                    cancelToken: _cancelTokenSource.Token);
+            case ApplicationState.Enabled when !isEnabled:
+            case ApplicationState.Disabled when isEnabled:
```
ludeeus, line 40:
```suggestion
            default:
```
It's a boolean so can only have 2 values?

FrankBakkerNl: the state is an enum that can be Enabled, disabled or some other states I think. But I agree the switch here could probably better be replaces with a normal if

ludeeus: lol, yeah I was about to comment 2 things;
1) This should probably be just a if
2) @ludeeus you have no clue what you are talking about 
:smile:

src/Runtime/NetDaemon.Runtime/Internal/AppStateManager.cs
```diff
@@ -1,67 +1,61 @@
...
+                await _appStateRepository.UpdateAsync(applicationId, isEnabled, _cancelTokenSource.Token)
+                    .ConfigureAwait(false);
                 break;
         }
     }
 
-    public void Initialize(IHomeAssistantConnection haConnection, IAppModelContext appContext)
+    public void Dispose()
```
FrankBakkerNl, line 47: Please make Initialize the first method after the constructor and dispose the last of the class

src/Runtime/NetDaemon.Runtime/Internal/AppStateRepository.cs
```diff
@@ -0,0 +1,70 @@
...
+        var entityId = EntityMapperHelper.ToSafeHomeAssistantEntityIdFromApplicationId(applicationId);
+
+        await haConnection.CallServiceAsync("input_boolean", enabled ? "turn_on" : "turn_off",
+            new HassTarget {EntityIds = new[] {entityId}},
+            cancelToken: token).ConfigureAwait(false);
+    }
+
+    public async Task RemoveNotUsedStatesAsync(IReadOnlyCollection<string> applicationIds, CancellationToken token)
+    {
+        if (applicationIds.Count == 0)
```
FrankBakkerNl, line 55: This seems odd, if there are no more apps you would expect all switches to be removed

src/Runtime/NetDaemon.Runtime/Internal/AppStateManager.cs
helto4real: Damn Rider does this every time :) I gonna see if there is a way to make it not do it!

src/Runtime/NetDaemon.Runtime/Internal/AppStateRepository.cs
helto4real: Indeed.

src/Runtime/NetDaemon.Runtime/Internal/AppStateRepository.cs
```diff
@@ -0,0 +1,70 @@
...
+            return;
+
+        var haConnection = _runner.CurrentConnection ??
+                           throw new InvalidOperationException();
+        var helpers = await haConnection.ListInputBooleanHelpersAsync(token).ConfigureAwait(false);
+
+        var entityIds = applicationIds.Select(EntityMapperHelper.ToSafeHomeAssistantEntityIdFromApplicationId);
+
+        var notUsedHelperIds = helpers.Where(n =>
```
FrankBakkerNl, line 64: This Contains() will execute the select() again for each helper causing each id to be mapped n^2 times. 
Instead of a where you could use Except() here that will use a hashTable interanlly

src/Runtime/NetDaemon.Runtime/Internal/AppStateRepository.cs
helto4real: You were right. I added test for it as well

src/Runtime/NetDaemon.Runtime/Internal/EntityMapperHelper.cs
```diff
@@ -0,0 +1,52 @@
...
+                case UnicodeCategory.SpaceSeparator:
+                case UnicodeCategory.ConnectorPunctuation:
+                case UnicodeCategory.DashPunctuation:
+                case UnicodeCategory.OtherPunctuation:
+                    if (lastChar != '_')
+                        stringBuilder.Append('_');
+                    break;
+            }
+
+            lastChar = c;
```
FrankBakkerNl, line 47: I think you should check if tha last appended char was an _, not the last input char. Now you could still get two _ in a row in the output

src/Runtime/NetDaemon.Runtime/Internal/NetDaemonRuntime.cs
```diff
@@ -97,12 +97,13 @@ CancellationToken cancelToken
 
             await _cacheManager.InitializeAsync(cancelToken).ConfigureAwait(false);
 
+
```
FrankBakkerNl, line 100: double newline

src/Runtime/NetDaemon.Runtime/Internal/AppStateRepository.cs
helto4real: The two lists are of two different types and I need the type that are the class to get the original Id back. It is such small lists so it will have no performance issues but will make it more complicated. If you have a suggestion please let me know but I cannot see it :)

FrankBakkerNl (changes requested):
finishe review, just some minors

src/Runtime/NetDaemon.Runtime/Internal/AppStateRepository.cs
FrankBakkerNl: Maybe Except is not easy here, then yoy could also just add ToHashSet at line 59

src/Runtime/NetDaemon.Runtime/Internal/EntityMapperHelper.cs
helto4real: I do not know any way to get last character added from a StringBuilder. But I think this logic still works.

src/Runtime/NetDaemon.Runtime/Internal/AppStateManager.cs
helto4real: Haha

src/Runtime/NetDaemon.Runtime/Internal/AppStateRepository.cs
helto4real: Good suggestion. Done

FrankBakkerNl (approved):

PR #658, opened 2022-02-04
Mqtt extension to create an update entities in HA

eugeneniemand:
## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist


src/Extensions/NetDaemon.Extensions.MqttEntityManager/MqttConfiguration.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace NetDaemon.Extensions.MqttEntityManager;
+
+public class MqttConfiguration
```
BeeHiveJava, line 3: What about TLS support?

BeeHiveJava: https://www.home-assistant.io/docs/mqtt/certificate/

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MqttEntityManager.cs
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    private static string ConfigPath(string domain, string entityId)
+    {
+        return $"{RootPath(domain, entityId)}/config";
+    }
+
+    private static string RootPath(string domain, string entityId)
+    {
+        return $"homeassistant/{domain}/{entityId}";
```
BeeHiveJava, line 55: This should be configurable: https://www.home-assistant.io/docs/mqtt/discovery/#discovery_prefix

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MqttConfiguration.cs
BeeHiveJava: Maybe we can add this at a later point, it's not blocking current functionality.

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MessageSender.cs
```diff
@@ -0,0 +1,64 @@
...
+        var connectResult = await mqttClient.ConnectAsync(options, CancellationToken.None).ConfigureAwait(false);
+        if (connectResult.ResultCode != MqttClientConnectResultCode.Success)
+            throw new AuthenticationException(connectResult.ReasonString);
+    }
+
+    private async Task PublishMessage(IApplicationMessagePublisher client, string topic, string payload)
+    {
+        var message = new MqttApplicationMessageBuilder().WithTopic(topic)
+                                                         .WithPayload(payload)
+                                                         .WithRetainFlag()
```
BeeHiveJava, line 55: Should this always retain? I think we should make that configurable as well. Retaining by default is sensible though.

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MessageSender.cs
```diff
@@ -0,0 +1,64 @@
...
+                                                         .Build();
+
+        _logger.LogDebug($"Sending to {message.Topic}: {message.ConvertPayloadToString()}");
```
BeeHiveJava, line 58: You can make use of structured logging here.

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MessageSender.cs
```diff
@@ -0,0 +1,64 @@
...
+        if (string.IsNullOrEmpty(_mqttConfig.Host))
+            throw new MqttConfigurationException("The Mqtt config was not found or there was an error loading it. Please add MqttConfiguration section to appsettings.json");
+
+        _logger.LogDebug($"MQTT connection is {_mqttConfig.Host}:{_mqttConfig.Port}/{_mqttConfig.UserId}");
+    }
+
+    public async Task SendMessageAsync(string topic, string payload)
+    {
+        using var mqttClient = _mqttFactory.CreateMqttClient();
+        await ConnectAsync(mqttClient);
```
BeeHiveJava, line 35: Is opening a connection every time a good idea? Should we pool the connection somewhere?

src/Extensions/NetDaemon.Extensions.MqttEntityManager/IMqttEntityManager.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace NetDaemon.Extensions.MqttEntityManager;
+
+public interface IMqttEntityManager
+{
+    Task CreateAsync(string domain, string deviceClass, string entityId, string name);
+    Task UpdateAsync(string domain, string entityId, string state, string? attributes = null);
+    Task RemoveAsync(string domain, string entityId);
```
BeeHiveJava, line 7: I'm wondering if we can make some kind of common value object here.

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MqttEntityManager.cs
BeeHiveJava: `homeassistant` Is the correct default.

src/Extensions/NetDaemon.Extensions.MqttEntityManager/IMqttEntityManager.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace NetDaemon.Extensions.MqttEntityManager;
+
+public interface IMqttEntityManager
+{
+    Task CreateAsync(string domain, string deviceClass, string entityId, string name);
```
BeeHiveJava, line 5: Would it be strange to swap around the `deviceClass` and `entityId` here? Then the starting arguments have the same position for every call here.

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MessageSender.cs
```diff
@@ -0,0 +1,64 @@
...
+    private readonly IMqttFactory           _mqttFactory;
+    private readonly MqttConfiguration      _mqttConfig;
+
+    public MessageSender(ILogger<MessageSender> logger, IConfiguration configuration, IMqttFactory mqttFactory, IOptions<MqttConfiguration> mqttConfig)
+    {
+        _logger      = logger;
+        _mqttFactory = mqttFactory;
+        _mqttConfig  = mqttConfig.Value;
+
+        if (string.IsNullOrEmpty(_mqttConfig.Host))
```
BeeHiveJava, line 26: This check is a little strange to me. Why is it only checking the host? And if the configuration section is missing from the appsettings, does it not throw an error somewhere?

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MessageSender.cs
skotl: Nope - it's supposed to be short lived (and actually has a default timeout of 10 seconds)

BeeHiveJava: Oh cool!

skotl: Actually, turns out that the MQTTNet project also supplies a managed mqtt client as an extension, which automatically reconnects on disconnection. It probably makes more sense to include that rather than the low-level client that is provided here.
I'll unresolve for now - I don't think it's blocking (as calls to mqtt will be relatively infrequent and it's not expensive to reconnect) but it's a useful improvement for later.

BeeHiveJava: That is also what I understood from the documentation. This already uses the managed client, it's just that it's disposed and recreated on every call.

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MqttEntityManager.cs
```diff
@@ -17,41 +20,41 @@ public async Task CreateAsync(string domain, string deviceClass, string entityId
...
     {
-        return $"{RootPath(domain, entityId)}/config";
+        return $"{RootPath(discoveryPrefix, domain, entityId)}/config";
     }
 
-    private static string RootPath(string domain, string entityId)
+    private static string RootPath(string discoveryPrefix, string domain, string entityId)
     {
-        return $"homeassistant/{domain}/{entityId}";
+        return $"{discoveryPrefix}/{domain}/{entityId}";
```
BeeHiveJava, line 85: Alternatively if you make this methods non-static, you don't have to pass the first argument and can get it from the options directly.

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MessageSender.cs
eugeneniemand: It actually still creates an instance of the class with all the default values

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MqttEntityManager.cs
eugeneniemand: Done

dev/DebugHost/apps/Extensions/MqttEntityManagerApp.cs
```diff
@@ -0,0 +1,63 @@
...
+
+    public MqttEntityManagerApp(IHaContext ha, ILogger<MqttEntityManagerApp> logger, IMqttEntityManager manager)
+    {
+        _ha      = ha;
+        _logger  = logger;
+        _manager = manager;
+    }
+
+    [SuppressMessage("Naming", "CA1727:Use PascalCase for named placeholders", Justification = "<Pending>")]
+    [SuppressMessage("Reliability", "CA2007:Consider calling ConfigureAwait on the awaited task", Justification = "<Pending>")]
```
helto4real, line 33: I do not think we should supress anything other than the CA1031 here

helto4real (changes requested):
I think we should use `ConfigureAwait(false)` and not supress those. Also keep only the interfaces used by users public.

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MessageSender.cs
```diff
@@ -0,0 +1,64 @@
...
+    public async Task SendMessageAsync(string topic, string payload)
+    {
+        using var mqttClient = _mqttFactory.CreateMqttClient();
+        await ConnectAsync(mqttClient);
+        await PublishMessage(mqttClient, topic, payload);
+    }
+
+    private async Task ConnectAsync(IMqttClient mqttClient)
+    {
+        var options = new MqttClientOptionsBuilder()
```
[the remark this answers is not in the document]
skotl: The Connect or the options builder?
This is the lightweight client whose connection is time sensitive (timeout is 10 seconds, I believe) so it's called when the message is ready to be sent. There is a managed client available as an extension that we could (should?) include, which maintains the connection.

dev/DebugHost/apps/Extensions/MqttEntityManagerApp.cs
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    [SuppressMessage("Naming", "CA1727:Use PascalCase for named placeholders", Justification = "<Pending>")]
+    public async Task InitializeAsync(CancellationToken cancellationToken)
+    {
+        try
+        {
+
+            _logger.LogInformation("Creating Entity {domain}.{entityId}", "binary_sensor", "manager_test");
+            await _manager.CreateAsync("binary_sensor", "motion", "manager_test", "Manager Test");
```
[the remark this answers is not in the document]
skotl: This is just the test app, which won't (or shouldn't!) be called by external apps - just a question of style really, but it's not a hardship to add this so we should do it.

src/Extensions/NetDaemon.Extensions.MqttEntityManager/IMessageSender.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace NetDaemon.Extensions.MqttEntityManager;
+
+public interface IMessageSender
```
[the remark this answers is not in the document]
skotl: It's public as it is in the constructor of the public `IMqttEntityManager` so that it can be supplied via DI.
Not sure how we work around that - do you have any suggestions?

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MessageSender.cs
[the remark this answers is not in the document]
skotl: Move options builder to constructor

dev/DebugHost/apps/Extensions/MqttEntityManagerApp.cs
```diff
@@ -0,0 +1,62 @@
...
+            await Task.Delay(250, cancellationToken).ConfigureAwait(false);
+            _logger.LogInformation("Entity {domain}.{entityId} State: {state} Attributes: {attributes}",
+                "binary_sensor", "manager_test", entity.State, entity.Attributes);
+
+            await _manager.RemoveAsync("binary_sensor", "manager_test").ConfigureAwait(false);
+            await Task.Delay(250, cancellationToken).ConfigureAwait(false);
+            var removed = _ha.Entity("binary_sensor.manager_test").State == null;
+            _logger.LogInformation("Removed Entity: {removed}", removed);
+        }
+        catch (Exception e)
```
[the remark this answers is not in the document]
skotl: This is just a catch-all to aid debugging in the test app - is that ok?

src/Extensions/NetDaemon.Extensions.MqttEntityManager/IMessageSender.cs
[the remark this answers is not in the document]
helto4real: It should be fine to make MqttManager internal. Then we only expose the IMqttManager for users.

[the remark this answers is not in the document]
skotl: Now internal, alongside `MqttEntityManager`

dev/DebugHost/apps/Extensions/MqttEntityManagerApp.cs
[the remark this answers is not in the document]
skotl: Done

helto4real (approved):
Good job @eugeneniemand and @skotl  really looking forward trying this!

PR #662, opened 2022-02-05
More optionality around entity creation, simplified public method signatures

skotl:
Provide more optionality around the create-entity requests.
Provide better defaults (unique_id, name) to ensure that a basic request to create an entity succeeds.
Change signatures of Create, Update and Remove to use a single `entityId` construct (like "sensor.kitchen_temp") rather than pass domain and identifier.
Expand test / sample application to show how to create a switch as well as different examples of optional arguments.

## Breaking change
There are two breaking changes - the requests to `CreateAsync`, `UpdateAsync` and `RemoveAsync` now require a single `entityId` argument (e.g. "sensor.kitchen_heat") rather than specifying a domain and identifier.
`UpdateAsync` no longer accepts a json string for attributes - if you have code that currently supplies this then the update will fail silently - supply a concrete or anonymous object to the method instead.

## Proposed change
Provide better default options to increase the chance of the creation succeeding.
Optional record parameter to `CreateAsync()` public method to allow caller to override each of these arguments.

Added more examples

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to documentation pull request: docs need to be updated - I'll do that

## Checklist

### DIFF
--- a/dev/DebugHost/apps/Extensions/MqttEntityManagerApp.cs
+++ b/dev/DebugHost/apps/Extensions/MqttEntityManagerApp.cs
@@ -15,7 +15,6 @@ using NetDaemon.HassModel;
 namespace DebugHost.apps.Extensions;
 
 [NetDaemonApp]
-[Focus]
 public class MqttEntityManagerApp : IAsyncInitializable
 {
     private readonly IHaContext _ha;
--- a/dev/DebugHost/apps/Extensions/MqttEntityManagerApp.cs
+++ b/dev/DebugHost/apps/Extensions/MqttEntityManagerApp.cs
@@ -50,8 +50,7 @@ public class MqttEntityManagerApp : IAsyncInitializable
                 .ConfigureAwait(false);
 
             // Change state of the new sensor, set an attribute to right now
-            await _entityManager.UpdateAsync("sensor.my_id", "shiny",
-                    JsonSerializer.Serialize(new { updated = DateTime.UtcNow }))
+            await _entityManager.UpdateAsync("sensor.my_id", "shiny", new { updated = DateTime.UtcNow })
                 .ConfigureAwait(false);
 
             // Walk through a more complete set of examples, checking HA to verify that each operation completed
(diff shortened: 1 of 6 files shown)


src/Extensions/NetDaemon.Extensions.MqttEntityManager/IMqttEntityManager.cs
```diff
@@ -8,17 +10,21 @@ public interface IMqttEntityManager
...
     /// <summary>
     ///     Update state and, optionally, attributes of an HA entity via MQTT
     /// </summary>
-    Task UpdateAsync(string domain, string entityId, string state, string? attributes = null);
+    Task UpdateAsync(string entityId, string state, string? attributes = null);
+    
+    /// <summary>
+    ///     Update state and, optionally, attributes of an HA entity via MQTT
+    /// </summary>
+    Task UpdateAsync(string entityId, string state, object? attributes = null);
```
eugeneniemand, line 29: This seems to be dupicated

eugeneniemand (approved):

src/Extensions/NetDaemon.Extensions.MqttEntityManager/IMqttEntityManager.cs
skotl: It's an overload to just supply an anonymous object rather than having to json.stringify the object

helto4real: I think we should only use the object? one. We should not make users use serializer to call an API.

dev/DebugHost/apps/Extensions/MqttEntityManagerApp.cs
```diff
@@ -18,15 +19,15 @@ namespace DebugHost.apps.Extensions;
 [Focus]
```
helto4real, line 19: Remove this before merge

helto4real (changes requested):
I think we should stick to the `object?` version only. This is how the API:s looks like in the rest of the NetDaemon product. We should not let users have to use JsonSerializers them self. Otherwise this looks good!

src/Extensions/NetDaemon.Extensions.MqttEntityManager/IMqttEntityManager.cs
skotl: Done

dev/DebugHost/apps/Extensions/MqttEntityManagerApp.cs
skotl: Oops - done

helto4real (approved):

PR #664, opened 2022-02-06
Mqtt managed client

skotl:
Migrate to managed MQTT client for more resilience

Note that this has a dependency on #662 

This is not a breaking change on its own but does incorporate breaking changes from #662

## Breaking change
Yes, via #662

## Proposed change
Move to MqttNet.Extensions.ManagedClient away from the simplified client
Wrap this up in an assurer class that the `MessageSender` class depends on for assuring a client is available (but is not necessarily connected).

## Type of change

- [x] Dependency upgrade - needs #662 
- [x] New feature (which adds functionality to an existing integration) - resilience

## Additional information

## Checklist

### DIFF
--- a/src/Extensions/NetDaemon.Extensions.MqttEntityManager/AssuredMqttConnection.cs
+++ b/src/Extensions/NetDaemon.Extensions.MqttEntityManager/AssuredMqttConnection.cs
@@ -1,0 +1,102 @@
+﻿using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using MQTTnet;
+using MQTTnet.Client.Connecting;
+using MQTTnet.Client.Disconnecting;
+using MQTTnet.Client.Options;
+using MQTTnet.Extensions.ManagedClient;
+using NetDaemon.Extensions.MqttEntityManager.Exceptions;
+
+namespace NetDaemon.Extensions.MqttEntityManager;
+
+/// <summary>
+/// Wrapper to assure an MQTT connection
+/// </summary>
+internal class AssuredMqttConnection : IAssuredMqttConnection, IDisposable
+{
+    private const int RetryMaximumSeconds = 15;
+    private readonly ILogger<AssuredMqttConnection> _logger;
+    private readonly Task? _connectionTask;
+    private IManagedMqttClient? _mqttClient;
+    private bool _disposed;
+
+    /// <summary>
+    /// Wrapper to assure an MQTT connection
+    /// </summary>
+    /// <param name="logger"></param>
+    /// <param name="mqttFactory"></param>
+    /// <param name="mqttConfig"></param>
+    public AssuredMqttConnection(
+        ILogger<AssuredMqttConnection> logger,
+        IMqttFactory mqttFactory,
+        IOptions<MqttConfiguration> mqttConfig)
+    {
+        _logger = logger;
+
+        _connectionTask = Task.Run(() => ConnectAsync(mqttConfig.Value, mqttFactory));
+    }
+
+    /// <summary>
+    /// Ensures that an MQTT client is available, retrying if necessary, and throws if connection is impossible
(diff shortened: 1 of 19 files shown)


src/Extensions/NetDaemon.Extensions.MqttEntityManager/MessageSender.cs
```diff
@@ -20,75 +18,44 @@ namespace NetDaemon.Extensions.MqttEntityManager;
...
+    ///     Publish a message to the given topic
     /// </summary>
     /// <param name="topic"></param>
     /// <param name="payload">Json structure of payload</param>
     /// <param name="retain"></param>
     public async Task SendMessageAsync(string topic, string payload, bool retain = false)
     {
-        using var mqttClient = _mqttFactory.CreateMqttClient();
-        await ConnectAsync(mqttClient);
+        var mqttClient = _assuredMqttConnection.GetClientOrThrow();
```
helto4real, line 42: This should be async. Then you could await it with a CancellationToken that are time out for example. Now you mix sync/async code and that often leads to unpredictable results like blocking. You could for example provide the CancellationToken to this method and that should be used while waiting for connection.

src/Extensions/NetDaemon.Extensions.MqttEntityManager/AssuredMqttConnection.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        _connectionTask = Task.Run(() => ConnectAsync(mqttConfig.Value, mqttFactory));
+    }
+
+    /// <summary>
+    /// Ensures that an MQTT client is available, retrying if necessary, and throws if connection is impossible
+    /// </summary>
+    /// <returns></returns>
+    /// <exception cref="MqttConnectionException">Timed out while waiting for connection</exception>
+    public IManagedMqttClient GetClientOrThrow()
```
helto4real, line 44: As comment before. This method can be async. Provide a CancellationToken and use ConnectAsync here using the cancellation token for cancel the potentially long running task connecting. I am sure the Mqtt client has a way to provide a cancellation token to the connect method.

src/Extensions/NetDaemon.Extensions.MqttEntityManager/AssuredMqttConnection.cs
```diff
@@ -0,0 +1,102 @@
...
+    {
+        _connectionTask?.Wait();
+
+        var retryCount = 1;
+        while ((_mqttClient == null || !_mqttClient.IsConnected) && retryCount <= RetryMaximumSeconds)
+        {
+            _logger.LogDebug("MQTT client is not yet available, retry #{RetryCount}", retryCount);
+            Thread.Sleep(1000);
```
helto4real, line 52: Using Thread.Sleep in a method that are not async but called from the async code could halt processing and create deadlocks.

helto4real (changes requested):

helto4real (changes requested):
The way of how async/sync code is written will be trouble I think. We should be able to make the GetClientOrThrow async.

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MessageSender.cs
skotl: I don't think we need the cancellationtoken in the new approach now, as the managed MQTT client will have its own timeout mechanism - would you agree?

src/Extensions/NetDaemon.Extensions.MqttEntityManager/AssuredMqttConnection.cs
skotl: As previous comment

helto4real (approved):
Looks way better now. Good job.

helto4real:
Closing this, already merged.

PR #677, opened 2022-02-11
Support dynamic options on MQTT entity creation

skotl:
Support dynamic options on MQTT entity creation
 1. Three new concrete config settings added: `QualityOfService`, `AvailabilityTopic` and dynamic `AdditionalOptions`
 1. `AvailabilityTopic` (like the other topic strings) cannot be overridden, but the other two can both be specified in the `EntityCreationOptions`
 1. `AvailabilityTopic` is only set when both the `payload_available` and `payload_not_available` options are supplied
 1. Dynamic `AdditionalOptions` is an optional set of config that can be supplied when creating an entity. The strongly-typed options are merged with the dynamic options (dynamic wins) and the combination is transmitted as the payload when creating an entity.
 1. `UpdateAsync()` now accepts an object as the new state so there is no need to `ToString`ify a value before calling
 1. New `SetAvailabilityAsync()` method added
 

## Breaking change
This is not a breaking change

## Proposed change
The current concrete entity creation configuration type is slightly extended but is now likely to stay static. It has an optional parameter, `AdditionalOptions`, which accepts a dynamic / anonymous (or any kind of) object and merges its properties with those of the concrete options.

The merge is done at the json level and properties specified in `AdditionalOptions` will override any property with the same name in concrete options.

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

Entity Manager documentation needs to be updated - I'll do this


src/Extensions/NetDaemon.Extensions.MqttEntityManager/Models/EntityCreationOptions.cs
```diff
@@ -1,15 +1,19 @@
...
+/// <param name="QualityOfService">Optional MQTT QOS level, default is 0</param>
+/// <param name="AdditionalOptions">Optional additional parameters to set</param>
 public record EntityCreationOptions(
     string? DeviceClass = null,
     string? UniqueId = null,
     string? Name = null,
-    bool Persist = true
+    bool Persist = true,
+    int QualityOfService = 0,
+    dynamic? AdditionalOptions = null
```
FrankBakkerNl, line 18: Use object iso dynamic

skotl: I want to translate it into a dynamic later on so that I can query its properties for merging with json structs, but that of course can be done easily by casting object to dynamic in the internal function.
I've verified that passing an anonymous object ( `new {...}`) when calling this casts without issue to the dynamic so I guess my question is why we would want to hide the dynamic type here.
Happy to change it, but I'm just keen to understand the benefit of object rather than dynamic in the signature.

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MqttEntityManager.cs
```diff
@@ -68,17 +59,55 @@ public async Task RemoveAsync(string entityId)
...
+    {
+        var (domain, identifier) = EntityIdParser.Extract(entityId);
+
+        await _messageSender.SendMessageAsync(AvailabilityPath(domain, identifier), availability).ConfigureAwait(false);
+    }
+
+    private string BuildCreationPayload(string domain, string identifier, string configPath,
+        EntityCreationOptions? options)
+    {
+        var availabilityRequired = // If payloads for availability are specified then we need the topic
```
FrankBakkerNl, line 94: Not sure I understand this, but it seems like this should be || instead f &&
If you make these payload_available and payload_not_available int? props on EntityCreationOptions this would be a simpler check

src/Extensions/NetDaemon.Extensions.MqttEntityManager/Helpers/EntityCreationPayloadHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// <returns></returns>
+    /// <exception cref="JsonException"></exception>
+    internal static string Merge(EntityCreationPayload concreteOptions, dynamic? dynamicOptions)
+    {
+        var concreteJson = JsonSerializer.SerializeToNode(concreteOptions)?.AsObject()
+                           ?? throw new JsonException("Unable to convert concrete config to JsonObject");
+
+        if (dynamicOptions != null)
+        {
+            JsonObject dynamicJson = JsonSerializer.SerializeToNode(dynamicOptions);
```
FrankBakkerNl, line 26: This line has a hidden cast from JsonNode to JsonObject. Because dynamicOptions is `dynamic` the compiler does not complain. I think it will actually break runtime if dynamicOptions would be an Array, because then SerializeToNode does not return a JsonObject. 

If you change dynamic to object here the compiler will complain and force you to cast explicitly like: 
```
             if (JsonSerializer.SerializeToNode(dynamicOptions) is JsonObject jsonObject)
                concreteJson.AddRange(jsonObject);
```

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MqttEntityManager.cs
```diff
@@ -68,17 +59,55 @@ public async Task RemoveAsync(string entityId)
...
     /// <param name="entityId"></param>
-    /// <param name="state">New state</param>
+    /// <param name="stateObject">New state, which will be converted to a string before submitting</param>
     /// <param name="attributes">Concrete or anonymous attributes</param>
-    public async Task UpdateAsync(string entityId, string? state, object? attributes = null)
+    public async Task UpdateAsync(string entityId, object? stateObject, object? attributes = null)
     {
         var (domain, identifier) = EntityIdParser.Extract(entityId);
-        
+        var state = stateObject?.ToString();
```
FrankBakkerNl, line 67: This object.ToString() could be culture specific in which case decimals or dates gets formatted incorrect. Internally in HA all states are string so maybe we should just leave this up to the client

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MqttEntityManager.cs
```diff
@@ -68,17 +59,55 @@ public async Task RemoveAsync(string entityId)
...
+
         if (!string.IsNullOrWhiteSpace(state))
```
FrankBakkerNl, line 69: I think it should be possible to set the state to an empty string. Not sure if it should also be possbile to set it to null to make it explicitly unavalable

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MqttEntityManager.cs
```diff
@@ -68,17 +59,55 @@ public async Task RemoveAsync(string entityId)
...
             await _messageSender.SendMessageAsync(StatePath(domain, identifier), state).ConfigureAwait(false);
-        
+
         if (attributes != null)
-            await _messageSender.SendMessageAsync(AttrsPath(domain, identifier), JsonSerializer.Serialize(attributes) ).ConfigureAwait(false);
+            await _messageSender.SendMessageAsync(AttrsPath(domain, identifier), JsonSerializer.Serialize(attributes))
```
FrankBakkerNl, line 73: Is it not possible maybe to set state and attributes in a single messsage? Maybe not because the way mqqt works. I just wonder if this also means it will create two state_change events in HA which would be unfortunate

src/Extensions/NetDaemon.Extensions.MqttEntityManager/Models/EntityCreationOptions.cs
```diff
@@ -1,15 +1,19 @@
...
 /// <param name="Name">Optional name of the entity</param>
 /// <param name="Persist">Optionally persist the entity over HA restarts, default is true</param>
+/// <param name="QualityOfService">Optional MQTT QOS level, default is 0</param>
+/// <param name="AdditionalOptions">Optional additional parameters to set</param>
 public record EntityCreationOptions(
     string? DeviceClass = null,
     string? UniqueId = null,
     string? Name = null,
-    bool Persist = true
+    bool Persist = true,
```
FrankBakkerNl, line 16: I cant find where this bool is used

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MqttEntityManager.cs
skotl: On the || vs &&, my reasoning was that setting availability without defining payloads for both would be pointless.
Hah! Having typed that, I've realised that someone might want to override one value but not the other - you're right: I'll change it to || 

On the concrete int? props question, I started a discussion on discord on whether these should be included in the concrete options _however_ they should be `string?` not `int?` because it is the value that the payload is compared to when they call the availability topic. So examples might be "down"/"up", "set", "unset", etc.

src/Extensions/NetDaemon.Extensions.MqttEntityManager/Helpers/EntityCreationPayloadHelper.cs
skotl: I'll try that - the decoding of a JsonNode is pretty fragile, to be honest, so I wouldn't be surprised if there were cases that caused it to break.
A more brutal solution might be to simply supply a `Dictionary<string, object>` for the optional params?

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MqttEntityManager.cs
skotl: Indeed. It is not possible, and I had a discussion with Eugene about this as I feared that we might be suggesting to the caller that this is a single atomic request.
GIven we now have the SetAvailability() request, I'm thinking we separate this method to `SetState()` and `SetAttributes()`

src/Extensions/NetDaemon.Extensions.MqttEntityManager/Helpers/EntityCreationPayloadHelper.cs
FrankBakkerNl: I think if the parameter is of type object the user actually can put a dictionary in that and this code would also work as expected because the serializer threats dictionaries special

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MqttEntityManager.cs
FrankBakkerNl: I think it would be safer indeed. Alternatieve might be to have specific overloads per type, or use a smarter ToString. Convert.ToString(state, invariantCulture( maybe. But there could be edge cases we did not think of like dates needing some special format etc. so string would be safest

src/Extensions/NetDaemon.Extensions.MqttEntityManager/Models/EntityCreationOptions.cs
skotl: Done - all references to `dynamic?` are now `object?`

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MqttEntityManager.cs
skotl: Addressed. The payloads for availability are now first class config options, and the availability topic address will be generated if either or both are supplied in the config

src/Extensions/NetDaemon.Extensions.MqttEntityManager/Helpers/EntityCreationPayloadHelper.cs
skotl: Done

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MqttEntityManager.cs
skotl: Reverted to string (and it's now in its own `SetStateAsync()` method)

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MqttEntityManager.cs
skotl: Now accepts empty string. I've not made the arg nullable as I think empty should suffice

src/Extensions/NetDaemon.Extensions.MqttEntityManager/MqttEntityManager.cs
skotl: I've replaced `UpdateAsync()` with `SetStateAsync()` and `SetAttributesAsync()` which seems cleaner

src/Extensions/NetDaemon.Extensions.MqttEntityManager/Models/EntityCreationOptions.cs
skotl: This (persist flags) was broken all over the place (retain was always being set), so now addressed as a property on the `MessageSender`, which honours the property when sending a message

src/Extensions/NetDaemon.Extensions.MqttEntityManager/Helpers/EntityCreationPayloadHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// <returns></returns>
+    /// <exception cref="JsonException"></exception>
+    internal static string Merge(EntityCreationPayload concreteOptions, object? additionalOptions)
+    {
+        var concreteJson = JsonSerializer.SerializeToNode(concreteOptions)?.AsObject()
+                           ?? throw new JsonException("Unable to convert concrete config to JsonObject");
+
+        if (additionalOptions != null)
+        {
+            JsonObject? dynamicJson = (JsonObject)JsonSerializer.SerializeToNode(additionalOptions)!;
```
FrankBakkerNl, line 26: Just to be safe: 

`JsonObject? dynamicJson = JsonSerializer.SerializeToNode(additionalOptions) as JsonObject;`

src/Extensions/NetDaemon.Extensions.MqttEntityManager/Models/EntityCreationOptions.cs
```diff
@@ -1,15 +1,19 @@
 ﻿namespace NetDaemon.Extensions.MqttEntityManager.Models;
```
FrankBakkerNl, line 1: Maybe this should be just in the same namespace as the interface, so the client does not need to add two usings

src/Extensions/NetDaemon.Extensions.MqttEntityManager/Helpers/EntityCreationPayloadHelper.cs
skotl: done

src/Extensions/NetDaemon.Extensions.MqttEntityManager/Models/EntityCreationOptions.cs
skotl: agree - done

FrankBakkerNl (approved):

PR #699, opened 2022-04-09
Support labeled selects in CodeGenerator

pathartl:
## Proposed change
Some option enumerables returned from HA will present options as label/value pairs starting with 2022.4 instead of just a string[] value. I added a new LabeledSelectSelector that will extract the string values. Note that not all selects in 2022.4 return as label/value pairs, some are still string[] and JSON deserialization has been altered to reflect this.

## Type of change

- [ ] Dependency upgrade
- [X] Bugfix (non-breaking change which fixes an issue)

## Checklist


helto4real:
As discussed in Discord group we need a long-term solution. Removing the options class will not be breaking since it is not used. A separate PR #701 fixes that.

PR #733, opened 2022-07-22
SubscribeSafe

helto4real:
## Breaking change

Changes behaviour of SubscribeAsync and SubscribeAsyncConcurrent to safely recover from unhandled exception. This was the intended behaviour from the beginning so technically it is a bugfix.

## Proposed change

This PR adds a safe subscription. Safe means it will recover from unhandled exceptions and continue process messages. Normal Subscribe would unsubscribe all subscribers when unhandled exceptions occurs.

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/src/HassModel/NetDeamon.HassModel/ObservableExtensions.cs
+++ b/src/HassModel/NetDeamon.HassModel/ObservableExtensions.cs
@@ -13,7 +13,20 @@ public static class ObservableExtensions
         Action<Exception>? onError = null)
     {
         return source
-            .Select(e => Observable.FromAsync(() => HandleTaskSafeAsync(onNextAsync, e, onError)))
+            .Select(e => Observable.FromAsync(() => HandleTaskSafeAsync(onNextAsync, e, onError, null)))
+            .Concat()
+            .Subscribe();
+    }
+
+    /// <summary>
+    ///     Allows calling async function but does this in a serial way. Long running tasks will block
+    ///     other subscriptions
+    /// </summary>
+    public static IDisposable SubscribeAsync<T>(this IObservable<T> source, Func<T, Task> onNextAsync,
+        ILogger logger)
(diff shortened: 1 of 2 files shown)


src/HassModel/NetDeamon.HassModel/ObservableExtensions.cs
```diff
@@ -69,7 +106,7 @@ private static T HandleTaskSafe<T>(Action<T> onNext, T e, Action<Exception>? onE
             else
             {
                 using var loggerFactory = LoggerFactory.Create(x => { x.AddConsole(); });
```
BeeHiveJava, line 108: Would be nice to only create the loggerFactory if it's needed - if you pass a logger, you won't need to create it.

skotl: Agree, also the error is only logged when the errorhandler is null, which can create confusion: if I passed an error handler _and_  a logger, I'd expect both to be honoured but in this case only the error handler would be.

I must admit to being unsure on the exact logic but perhaps it's like this?

1. If an error handler is passed, call it
2. If a logger is passed, call it
3. If neither, then create a default logger and log a generic message

1.

helto4real: I did add overload method just to pass a logger now.  @BeeHiveJava 
1. You not provide anything _> use default factory
2. You provide logger, use that
3. You provee callback, use that.
They cannot be used in combination.

helto4real: Ohh I realize now I forgot to add that logic :) haha good eye!

helto4real:
Merging this due to low activity on reviews

Issue #685, opened 2022-02-20
Generated code looses attributes that are not present when running the nd-codgen tool.


FrankBakkerNl:
## The problem

The generated code has a class with properties that match the attributes for tje entities in each domain. Some entities will not always have all attributes present. eg a light entity only has a brightness attribute when it is turned on. As a result when re-generating the code while all lights are off the generated LightEntityEttributes class will not have a Brightness property. Any used code that uses this property will then fail to compile.
## Environment

22.6.3

## Traceback/Error logs

```txt

```

helto4real:
Yea we might need to handle this by saving attributes in some file so we at least keep attributes discovered in previous scans

FrankBakkerNl:
I have a draft PR up for this that will solve the most common cases (lights off during codegen). What specific entities / attributes did you have problems with?

Tsuy89:
Good to hear there is something in the pipeline! My case was also the brightness attribute, same as the example.

PR #762, opened 2022-10-08
Added MediaPlayerAttributesBase Class & Fixed Types in LightAttributesBase Class

phillipsawyer:
## Proposed change

Added the MediaPlayerAttributesBase class which was very simple and didn't require any other code modification. This means that I did not add any additional tests 

I partly used https://developers.home-assistant.io/docs/core/entity/media-player for reference, but excluded the `media_image_remotely_accessible` field since it never appears in the forums so it's too rare to be worth crowding the autocomplete list.

I didn't add any tests since I didn't change any code other than add the baseclass and it is working perfectly through all my testing & the existing tests all pass.

## Type of change

- [ ] Dependency upgrade
- [X] New feature (which adds functionality to an existing integration)

## Additional information

- This PR fixes or closes issue: 
- This PR is related to issue: https://github.com/net-daemon/netdaemon/issues/685

## Checklist

### DIFF
--- a/src/HassModel/NetDeamon.HassModel/Entities/Core/LightAttributesBase.cs
+++ b/src/HassModel/NetDeamon.HassModel/Entities/Core/LightAttributesBase.cs
@@ -3,9 +3,6 @@
 
 public record LightAttributesBase
 {
-    [JsonPropertyName("entity_id")]
-    public IReadOnlyList<string>? EntityId { get; init; }
-
     /// <summary>
     /// Integer between 0 and 255 for how bright the light should be, where 0 means the light is off, 1 is the minimum brightness and 255 is the maximum brightness supported by the light.
     /// </summary>
@@ -14,21 +11,42 @@ public record LightAttributesBase
     
     [JsonPropertyName("color_mode")]
     public string? ColorMode { get; init; }
+    
+    /// <summary>
+    /// Integer in mireds representing the color temperature of the light.
+    /// </summary>
+    [JsonPropertyName("color_temp")]
+    public int? ColorTemp { get; init; }
 
     /// <summary>
     /// The current effect.
     /// </summary>
     [JsonPropertyName("effect")]
     public string? Effect { get; init; }
-    
+
     /// <summary>
     /// The list of supported effects.
     /// </summary>
     [JsonPropertyName("effect_list")]
     public IReadOnlyList<string>? EffectList { get; init; }
+    
+    /// <summary>
+    /// Entity ids of the entities in the light group. Null if not a group.
+    /// </summary>
+    [JsonPropertyName("entity_id")]
+    public IReadOnlyList<string>? EntityId { get; init; }
 
+    /// <summary>
+    /// Name of the light as displayed in the UI.
+    /// </summary>
     [JsonPropertyName("friendly_name")]
     public string? FriendlyName { get; init; }
+    
+    /// <summary>
+    /// List containing two floats representing the hue and saturation of the color of the light. Hue is scaled 0-360, and saturation is scaled 0-100.
+    /// </summary>
+    [JsonPropertyName("hs_color")]
+    public IReadOnlyList<double>? HsColor { get; init; }
 
     [JsonPropertyName("icon")]
     public string? Icon { get; init; }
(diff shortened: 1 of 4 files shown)


src/HassModel/NetDeamon.HassModel/Entities/Core/LightAttributesBase.cs
```diff
@@ -3,42 +3,90 @@
...
+    [JsonPropertyName("off_brightness")]
+    public object? OffBrightness { get; init; }
+    
     [JsonPropertyName("supported_color_modes")]
     public IReadOnlyList<string>? SupportedColorModes { get; init; }
-
+    
     [JsonPropertyName("supported_features")]
-    public double? SupportedFeatures { get; init; }
+    public int? SupportedFeatures { get; init; }
```
helto4real, line 55: The move from double to inte is a breaking change but it seem more correct using ints. @FrankBakkerNl is this the old behaviour of codegen before we had base class to generate doubles? I am not sure why I added them as doubles... hmm

src/HassModel/NetDeamon.HassModel/Entities/Core/LightAttributesBase.cs
```diff
@@ -3,42 +3,90 @@
...
     
+    /// <summary>
+    /// Integer in mireds representing the color temperature of the light.
+    /// </summary>
     [JsonPropertyName("color_temp")]
-    public IReadOnlyList<double>? ColorTemp { get; init; }
```
helto4real, line 34: Ohhh the color temp was a list? Hmm that does not seem right. This is a bug I guess. It should be corrected

helto4real (commented):
I think it looks good. I want a discussion with @FrankBakkerNl about the changes from double to int. I wait for him to respond here or I will ping him on discord tomorrow and we discuss there

src/HassModel/NetDeamon.HassModel/Entities/Core/LightAttributesBase.cs
FrankBakkerNl: TLDR: I agree


We used double for generated properties because we can't know from sample data if it is an int or a double. In this case when hard coding we do know it to be an int so this makes sense to me. It could be breaking in theory but it is unlikely ppl actually use this and depend on it being a double

src/HassModel/NetDeamon.HassModel/Entities/Core/LightAttributesBase.cs
FrankBakkerNl: If I recall correctly, I copied these properties originally from my own generated code. If that is true it was an array at some point. If you are sure it's not this change is ok by me

phillipsawyer: Definitely serialized as a `double?` for me by the code gen, I just assumed you accidentally copied the same format as the other color properties. You also left the `EntityId` as `object?` instead of `IReadOnlyList<string>?` which I just corrected with another commit. Assuming there was no reason you put it in as an `object?` instead.

src/HassModel/NetDeamon.HassModel/Entities/Core/LightAttributesBase.cs
phillipsawyer: Good point! Wasn't thinking much of it nor as a breaking change while I was doing it bc of implicit cast.

phillipsawyer: It should probably be a `long` instead of an `int` eh? Just changed it to longs and found one more case where double -> long (SupportedFeatures in MediaPlayer)

src/HassModel/NetDeamon.HassModel/Entities/Core/MediaPlayerAttributesBase.cs
```diff
@@ -17,7 +17,7 @@ public record MediaPlayerAttributesBase
     public string? DeviceClass { get; init; }
 
     [JsonPropertyName("entity_id")]
-    public object? EntityId { get; init; }
+    public IReadOnlyList<string>? EntityId { get; init; }
```
FrankBakkerNl, line 23: Same here, I prefer to be safe in these cases because it will break at runtime

FrankBakkerNl (approved):

phillipsawyer:
I'm pretty confident about them always being `IReadOnlyList<string>` in these two cases. If someone comes up with an edge case, I'd be happy to write a custom serializer to convert `string` to a 1 long `IReadOnlyList<string>` in those rare cases.

PR #767, opened 2022-10-23
Code gen merge previous and current metadata

### LINKED ISSUE #685: Generated code looses attributes that are not present when running the nd-codgen tool.

FrankBakkerNl:
## Proposed change

Fixed a log standing issue with properties disappearing in generated code when no entity currently provides them. eg a Light does not have a `brightness` property when it is off.

The code generator now saves the metadata it reads from Home Assistant to disk in Json files. It will then use these file on subsequent runs and merge the current entity metadata from HA with the previous data loaded from these files

## Type of change

- [ ] Dependency upgrade
- [ X] Bugfix (non-breaking change which fixes an issue)

## Additional information

- This PR fixes or closes issue: fixes #685 

## Checklist


helto4real (approved):
Looks good. Very nice refactoring too.

PR #789, opened 2022-11-25
Add timeout to AssuredMqttConnection

skotl:
Address potential stall scenario where AssuredMqttConnection could wait forever / a very long time to connect to MQTT.

Forced timeout of 30 seconds and added unit test to exercise the scenario.

## Non-breaking change
..._except_... in a situation where connecting to MQTT might take more than 30 seconds (which seems unlikely)

## Proposed change
Every request to the MQTT Entity Manager makes a call to the underlying `AssuredMqttConnection` class. This class sets up a reliable connection to MQTT which includes auto-reconnect.
The purpose of this class is to ensure that before we execute a fire-and-forget request to MQTT that the MQTT connection is actually in place.

Theoretically, this should be fool-proof (aha!) in that the auto-reconnect will allow for network dropouts. However a user reported that he is seeing stalls somewhere within MQTT and, as the only blocking call in the stack, this seems the most likely culprit.

So this change wraps a 30 second timeout around the return of the connection. An important point is that once a connection to a live MQTT bus is returned then this is effectively self-maintaining, with the underlying `ManagedMqttClient` handling its own reconnection to dropped destinations.

In other words, this change will abort a connection to a non-responsive MQTT but will not impact behaviour after that initial connection.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes an issue reported in Discord

## Checklist

No user-visible changes, no documentation update needed

### DIFF
--- a/src/Client/NetDaemon.HassClient.Tests/ExtensionsTest/MqttEntityManagerTests/AssuredMqttConnectionTests.cs
+++ b/src/Client/NetDaemon.HassClient.Tests/ExtensionsTest/MqttEntityManagerTests/AssuredMqttConnectionTests.cs
@@ -15,7 +15,7 @@ public class AssuredMqttConnectionTests
         var mqttFactory = new MqttFactoryWrapper(mqttClient.Object);
        
         var conn = new AssuredMqttConnection(logger.Object, mqttFactory, GetMockOptions());
-        var returnedClient = await conn.GetClientAsync(30);
+        var returnedClient = await conn.GetClientAsync();
         
         returnedClient.Should().Be(mqttClient.Object);
     }
@@ -23,17 +23,17 @@ public class AssuredMqttConnectionTests
     [Fact]
     public async Task GetClientAbortsAfterTimeout()
     {
-        const int timeoutSeconds = 1;
+        var timeout = new TimeSpan(0, 0, 0, 0, 10);
         var logger = new Mock<ILogger<AssuredMqttConnection>>();
         
         var mqttClient = new Mock<IManagedMqttClient>();
         var mqttFactory = new MqttFactoryWrapper(mqttClient.Object);
 
         mqttClient.Setup(m => m.StartAsync(It.IsAny<ManagedMqttClientOptions>()))
-            .Callback(() => Thread.Sleep(timeoutSeconds * 2000));
+            .Callback(() => Thread.Sleep(timeout*2));
         
         var conn = new AssuredMqttConnection(logger.Object, mqttFactory, GetMockOptions());
-        var act = async () => { await conn.GetClientAsync(timeoutSeconds); };
+        var act = async () => { await conn.GetClientAsync(timeout); };
         
         await act.Should().ThrowAsync<TimeoutException>();
     }
--- a/src/Client/NetDaemon.HassClient.Tests/ExtensionsTest/MqttEntityManagerTests/AssuredMqttConnectionTests.cs
+++ b/src/Client/NetDaemon.HassClient.Tests/ExtensionsTest/MqttEntityManagerTests/AssuredMqttConnectionTests.cs
@@ -23,14 +23,14 @@ public class AssuredMqttConnectionTests
     [Fact]
     public async Task GetClientAbortsAfterTimeout()
     {
-        var timeout = new TimeSpan(0, 0, 0, 0, 10);
+        var timeout = TimeSpan.FromMilliseconds(5);
         var logger = new Mock<ILogger<AssuredMqttConnection>>();
         
         var mqttClient = new Mock<IManagedMqttClient>();
         var mqttFactory = new MqttFactoryWrapper(mqttClient.Object);
 
         mqttClient.Setup(m => m.StartAsync(It.IsAny<ManagedMqttClientOptions>()))
-            .Callback(() => Thread.Sleep(timeout*2));
+            .Callback(() => Thread.Sleep(TimeSpan.FromSeconds(1)));
         
         var conn = new AssuredMqttConnection(logger.Object, mqttFactory, GetMockOptions());
         var act = async () => { await conn.GetClientAsync(timeout); };
--- a/src/Client/NetDaemon.HassClient.Tests/ExtensionsTest/MqttEntityManagerTests/AssuredMqttConnectionTests.cs
+++ b/src/Client/NetDaemon.HassClient.Tests/ExtensionsTest/MqttEntityManagerTests/AssuredMqttConnectionTests.cs
@@ -15,12 +15,30 @@ public class AssuredMqttConnectionTests
         var mqttFactory = new MqttFactoryWrapper(mqttClient.Object);
        
         var conn = new AssuredMqttConnection(logger.Object, mqttFactory, GetMockOptions());
-        var returnedClient = await conn.GetClientAsync();
+        var returnedClient = await conn.GetClientAsync(30);
         
         returnedClient.Should().Be(mqttClient.Object);
     }
+
+    [Fact]
+    public async Task GetClientAbortsAfterTimeout()
+    {
+        const int timeoutSeconds = 1;
+        var logger = new Mock<ILogger<AssuredMqttConnection>>();
+        
+        var mqttClient = new Mock<IManagedMqttClient>();
+        var mqttFactory = new MqttFactoryWrapper(mqttClient.Object);
+
+        mqttClient.Setup(m => m.StartAsync(It.IsAny<ManagedMqttClientOptions>()))
+            .Callback(() => Thread.Sleep(timeoutSeconds * 2000));
+        
+        var conn = new AssuredMqttConnection(logger.Object, mqttFactory, GetMockOptions());
+        var act = async () => { await conn.GetClientAsync(timeoutSeconds); };
+        
+        await act.Should().ThrowAsync<TimeoutException>();
+    }
     
-    private IOptions<MqttConfiguration> GetMockOptions()
+    private static IOptions<MqttConfiguration> GetMockOptions()
     {
         var options = new Mock<IOptions<MqttConfiguration>>();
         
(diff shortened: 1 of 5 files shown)


src/Extensions/NetDaemon.Extensions.MqttEntityManager/AssuredMqttConnection.cs
```diff
@@ -38,18 +38,20 @@ internal class AssuredMqttConnection : IAssuredMqttConnection, IDisposable
     /// <summary>
     /// Ensures that the MQTT client is available
     /// </summary>
+    /// <param name="timeoutSeconds">Number of seconds to wait for connection to become available</param>
     /// <returns></returns>
     /// <exception cref="MqttConnectionException">Timed out while waiting for connection</exception>
-    public async Task<IManagedMqttClient> GetClientAsync()
+    public async Task<IManagedMqttClient> GetClientAsync(int timeoutSeconds)
```
FrankBakkerNl, line 44: Maybe use a timepan as an argument here?

src/Extensions/NetDaemon.Extensions.MqttEntityManager/AssuredMqttConnection.cs
```diff
@@ -38,18 +38,20 @@ internal class AssuredMqttConnection : IAssuredMqttConnection, IDisposable
...
     {
-        await _connectionTask;
+        if (await Task.WhenAny(_connectionTask, Task.Delay(timeoutSeconds * 1000)) == _connectionTask)
```
FrankBakkerNl, line 46: There is actually Task.WaitAsync(TimeSpan) in .Net 6

https://learn.microsoft.com/en-in/dotnet/api/system.threading.tasks.task.waitasync?view=net-6.0

src/Extensions/NetDaemon.Extensions.MqttEntityManager/AssuredMqttConnection.cs
FrankBakkerNl: Seems like a breaking change? Add an overload without a timeout

src/Client/NetDaemon.HassClient.Tests/ExtensionsTest/MqttEntityManagerTests/AssuredMqttConnectionTests.cs
```diff
@@ -15,12 +15,30 @@ public async Task CanGetClient()
...
+    public async Task GetClientAbortsAfterTimeout()
+    {
+        const int timeoutSeconds = 1;
+        var logger = new Mock<ILogger<AssuredMqttConnection>>();
+        
+        var mqttClient = new Mock<IManagedMqttClient>();
+        var mqttFactory = new MqttFactoryWrapper(mqttClient.Object);
+
+        mqttClient.Setup(m => m.StartAsync(It.IsAny<ManagedMqttClientOptions>()))
+            .Callback(() => Thread.Sleep(timeoutSeconds * 2000));
```
FrankBakkerNl, line 33: I don't really like Sleeps in test, This test will now always take 1 second to run.

If you change the api to take a timespan iso int you can make the timeout 10ms during the test.

src/Extensions/NetDaemon.Extensions.MqttEntityManager/AssuredMqttConnection.cs
skotl: Good shout - new commit with WaitAsync()

src/Extensions/NetDaemon.Extensions.MqttEntityManager/AssuredMqttConnection.cs
skotl: Could do - probably a style debate. I like the seemingly-unambiguous "timeoutSeconds" but I can see a TimeSpan being even less ambiguous.
Your call - happy to change it.

src/Client/NetDaemon.HassClient.Tests/ExtensionsTest/MqttEntityManagerTests/AssuredMqttConnectionTests.cs
skotl: Ah - makes sense. Ignore my previous comment on timespans.

Also agree that I don't like sleeps in tests - I had figured that the whole suite takes around 8 seconds on my setup but you're right; there are degrees of "badness" :)
Will update to TimeSpan

Do you have any other suggestions on asserting on time-passing?

FrankBakkerNl: Not really, unless you want to inject schedulers and mocked schedulers etc, which is a big overkill here.

src/Extensions/NetDaemon.Extensions.MqttEntityManager/AssuredMqttConnection.cs
```diff
@@ -38,18 +38,22 @@ internal class AssuredMqttConnection : IAssuredMqttConnection, IDisposable
...
+    public async Task<IManagedMqttClient> GetClientAsync(int timeoutSeconds)
     {
-        await _connectionTask;
+        await _connectionTask.WaitAsync(new TimeSpan(0, 0, timeoutSeconds));
+        
+        if (_connectionTask.IsCompletedSuccessfully)
+            return _mqttClient ?? throw new MqttConnectionException("Unable to create MQTT connection");
 
-        return _mqttClient ?? throw new MqttConnectionException("Unable to create MQTT connection");
+        throw new TimeoutException($"Failed to connect to MQTT broker after {timeoutSeconds} seconds");
```
FrankBakkerNl, line 51: Oops, you forgot to remove this

src/Extensions/NetDaemon.Extensions.MqttEntityManager/AssuredMqttConnection.cs
skotl: Scrap that - went with TimeSpan to allow a shorter duration in the tests

src/Client/NetDaemon.HassClient.Tests/ExtensionsTest/MqttEntityManagerTests/AssuredMqttConnectionTests.cs
skotl: Agreed - new test with timespan is in the ~20ms range

src/Extensions/NetDaemon.Extensions.MqttEntityManager/AssuredMqttConnection.cs
skotl: Nope - that's intentional. The first check captures some unforeseen issue in connecting to MQTT, the second throw is a direct consequence of the timeout.

skotl: (I think!)

FrankBakkerNl: The WaitAsync will already throw a TomeoutExecption, so if you get here the Original Task must be competed and the if is actually not needed as wel

skotl: Took me a moment but I get it now - good catch

FrankBakkerNl: I actualy think the _mqttClient field is not needed if the Task just returns the client

FrankBakkerNl: Then this entry method would be just

`return await _connectionTask.WaitAsync(new TimeSpan(0, 0, timeoutSeconds));'

FrankBakkerNl: and ConnectAsync returns a Task<IManagedMqttClient>

FrankBakkerNl: Maybe I am making things to complicated. But _connectionTask.IsCompletedSuccessfully cannot be false after you awaited the task

FrankBakkerNl: So at that point you know the _mqttClient will be initilaized.

You could just do 

return _mqttClient ?? throw new UnreachableException();


(Just learned abut UnreachableException today)

eugeneniemand:
I've checked out this PR, built locally and added the binary to my instance to see if I still experience these timeouts. I have a sensor that updates every 30 minutes and often get stuck within a few hours after a ND restart. On aside note you mentioned network dropouts, my ND and MQTT is on the same device, both running in Docker. I think I experienced this since we upgraded MQTT library to the newer version but that my just be coincidence. I'll keep you posted.

skotl:
> 

Cool - thanks. Let me know if I can help. Note that this will throw and log an error if it times out, but you probably still want to consider the more detailed logging in tandem.

eugeneniemand:
Ok this has now been running for 12 hours and I have no exceptions or errors, unless my logging levels are incorrect. However, Its not gotten stuck yet and still performs as expected

src/Extensions/NetDaemon.Extensions.MqttEntityManager/AssuredMqttConnection.cs
phillipsawyer: Agreed Frank, and thus the exception comment about "MqttConnectionException" being "Timed out while waiting for connection" is not correct. It also throws the TimeoutException now which is not in the method summary comment.

Minor picky details but timeout is now a timespan so it is not the "number of seconds".

helto4real:
Is this ready for merge? It's two month old now and I consider close it

FrankBakkerNl (commented):
fix waring

helto4real:
I will close this PR soon unless someone fix it to be mergable. It is not good to have stale PR:s. Thanks for understanding. @skotl

helto4real:
Stale...

PR #818, opened 2022-12-20
Add deviceId to Entity class and codegen

Ikcelaks:
## Proposed change

This PR adds the DeviceId property to the Entity class and updates the Code Generator to populate all entities with the DeviceId of their associated device.

A future enhancement could augment the existing EntityAreaCache code to also track the device registry, which is mutable in Home Assistant. Since the DeviceId associated with an entity will never change, it makes sense to bake it into the Entity class.

The immediate benefit is that users can now access the DeviceId of their devices via the DeviceId property of any of the associated entities. A specific use case is listening for the "zwave_js_value_notification" event for a specific device.

I have marked the PR as "draft" because it is as yet untested beyond running the test suite, and I've only added a single test so far. Before putting in the real work, I want to gather feedback on whether this enhancement acceptable before I put in the real testing work.

Side note: I considered whether it would be a good idea to add other properties such as AreaId to the codegen, but I think that only truly immutable properties should be baked in. An entity might change areas, get renamed, etc., but the DeviceId should never change.

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/src/HassModel/NetDaemon.HassModel.CodeGenerator/CodeGeneration/EntitiesGenerator.cs
+++ b/src/HassModel/NetDaemon.HassModel.CodeGenerator/CodeGeneration/EntitiesGenerator.cs
@@ -79,7 +79,7 @@ internal static class EntitiesGenerator
     {
         var entityName = EntityIdHelper.GetEntity(entity.id);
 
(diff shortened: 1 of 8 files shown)


Ikcelaks:
After chatting with @FrankBakkerNl , I'm closing this PR and will retry later with an attempt that works beyond only code generated entities.

helto4real:
Thank you for your contribution. Even not accepted PR:s moves the project forward.

PR #821, opened 2022-12-23
Add an entity model that support generic state types

### LINKED ISSUE #715: ProximityEntity state should be numeric
## Problem: State for ProximityEntity is a string
On the contrary, it is always measured in km, miles or other distance units. 

 ## Solution: Alter Code Generator
Use `NumericEntityState` to define `ProximityEntity`:

```csharp
public record ProximityEntity : Entity<ProximityEntity, NumericEntityState<ProximityAttributes>, ProximityAttributes>
```

## More Info

Reference: [Proximity Integration](https://www.home-assistant.io/integrations/proximity/)

Ikcelaks:
## Proposed change

This PR proposes a method for enabling full generics flexibility to the entity model. This is done by creating a new family of interfaces that can exist in parallel with the existing classes. This means there is no breaking change required!

The only compromise is that a couple of methods in `IHaContext` need to be duplicated so that the home assistant state can be accessed through a different type. Also, a few of the concrete class names are slightly off from ideal, but the user won't actually interact with them when writing their apps.

I'm submitting this as a draft to get early thoughts on naming and the general structure of the code. As of right now, the code compiles without warnings and passes the first batch of tests that I stole from the tests for the existing Entity code. These tests don't give full coverage, and I will write more and more specific tests. That said, I think the basic stuff will work, but I haven't written any Apps that use it yet.

As a basic guide, the user should NEVER interact directly with any of the concrete classes/records. All interaction is with `IEntity`, `IEntyState`, `IStateChange`, etc. Furthermore, the user will almost exclusively deal with the generic `<TState, TAttributes>` versions of the interfaces, so don't worry too much about all the `Raw*` properties. Users won't be dealing with them.

Here is a partial list of equivalents:
`Entity` is `IEntity<string?, Dictionary<string, object>>`
`Entity<TAttributes>` is `IEntity<string?, TAttributes>`
`NumericEntity` is `IEntity<double?, Dictionary<string, object>>`
`NumericEntity<TAttributes>` is `IEntity<double?, TAttributes>`
The `DateTime` entity type is `IEntity<DateTime?, TAttributes>`

You could make a `IEntity<LockStatus, TAttributes>` that mapped the state to a class that had properties like `IsLocked` etc. I just haven't done it yet, because others need to look at the code first.

One big question is about the default Attributes type. On the existing model, it's `object?`. Here, I just let it be the `Dictionary<string?, object>` type, but that should almost certainly change to something less implementation specific. Honestly, I have a hard time caring about it too much, because I wouldn't ever use it.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #715

## Checklist

### DIFF
--- a/src/HassModel/NetDeamon.HassModel/Entities/EntityGeneric.cs
+++ b/src/HassModel/NetDeamon.HassModel/Entities/EntityGeneric.cs
@@ -1,27 +1,35 @@
 namespace NetDaemon.HassModel.Entities;
 
-public record EntityGeneric : IEntity
+/// <inheritdoc/>
+public sealed record EntityGeneric : IEntity
 {
+    /// <inheritdoc/>
     public IHaContext HaContext { get; }
 
+    /// <inheritdoc/>
     public string EntityId { get; }
 
-    public EntityGeneric(IHaContext haContext, string entityId)
+    internal EntityGeneric(IHaContext haContext, string entityId)
     {
         HaContext = haContext;
         EntityId = entityId;
     }
 
+    /// <inheritdoc/>
     public IEntityState? RawEntityState => HaContext.GetStateGeneric(EntityId);
 
+    /// <inheritdoc/>
     public string? RawState => RawEntityState?.RawState;
 
+    /// <inheritdoc/>
     public IObservable<IStateChange> RawStateAllChanges() =>
         HaContext.StateAllChangesGeneric().Where(e => e.Entity.EntityId == EntityId);
 
+    /// <inheritdoc/>
     public IObservable<IStateChange> RawStateChanges() =>
         RawStateAllChanges().StateChangesOnlyGeneric();
 
+    /// <inheritdoc/>
     public void CallService(string service, object? data = null)
     {
         ArgumentNullException.ThrowIfNull(service, nameof(service));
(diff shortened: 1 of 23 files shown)


src/HassModel/NetDeamon.HassModel/IHaContext.cs
```diff
@@ -15,17 +15,36 @@ public interface IHaContext
     /// </summary>
     IObservable<StateChange> StateAllChanges();
 
+    /// <summary>
+    /// The observable state stream, all changes including attributes
+    /// </summary>
+    /// <returns></returns>
+    IObservable<IStateChange> StateAllChangesGeneric();
```
FrankBakkerNl, line 22: I think this could be an extension method, so you would not need to change this interface and implementation

src/HassModel/NetDeamon.HassModel/Entities/IEntity.cs
```diff
@@ -0,0 +1,100 @@
...
+    /// <summary>
+    /// Calls a service using this entity as the target
+    /// </summary>
+    /// <param name="service">Name of the service to call. If the Domain of the service is the same as the domain of the Entity it can be omitted</param>
+    /// <param name="data">Data to provide</param>
+    void CallService(string service, object? data = null);
+}
+
+/// <summary>Represents a Home Assistant entity with its state, changes and services</summary>
+public interface IEntity<TState, TAttributes> : IEntity
```
FrankBakkerNl, line 44: I think most type parameters on these interfaces can be marked `out` to make them covariant and make casting easier

src/HassModel/NetDeamon.HassModel/Entities/IEntity.cs
```diff
@@ -0,0 +1,100 @@
...
+    /// </summary>
+    IObservable<IStateChange<TState, TAttributes>> StateChanges();
+
+    /// <summary>
+    /// Get a new IEntity with new state type mapping
+    /// </summary>
+    /// <param name="newStateParser"></param>
+    /// <typeparam name="TStateNew"></typeparam>
+    /// <returns></returns>
+    IEntity<TStateNew, TAttributes> WithStateAs<TStateNew>(Func<string?, TStateNew> newStateParser);
```
FrankBakkerNl, line 82: I dont really see a usecase for this, maybe better to make it an extension method (same with the `WithAttributesAs<TAttributesNew>(Func<JsonElement?,..`)

src/HassModel/NetDeamon.HassModel/Entities/IEntity.cs
```diff
@@ -0,0 +1,100 @@
+namespace NetDaemon.HassModel.Entities;
+
+/// <summary>Represents a Home Assistant entity with its state, changes and services</summary>
+public interface IEntity
```
FrankBakkerNl, line 4: Is this non-generic interface actually needed

src/HassModel/NetDeamon.HassModel/Entities/IEntityFactory.cs
```diff
@@ -0,0 +1,15 @@
+namespace NetDaemon.HassModel.Entities;
+
+/// <summary>
+/// Factory for creating strongly typed generic IEntity instances
+/// </summary>
+public interface IEntityFactory
```
FrankBakkerNl, line 6: For the factory we might  need just a concrete class (without an interface)

src/HassModel/NetDeamon.HassModel/Entities/EntityGeneric.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        HaContext.CallService(serviceDomain, serviceName, ServiceTarget.FromEntity(EntityId), data);
+    }
+}
+
+/// <inheritdoc/>
+public sealed record EntityGeneric<TState, TAttributes> : IEntity<TState, TAttributes>
+    where TAttributes : class
+{
+    private readonly IEntity _entity;
```
FrankBakkerNl, line 49: So part of the trick here is to use  'containment over inheritance'

src/HassModel/NetDeamon.HassModel/Entities/IStateChange.cs
```diff
@@ -0,0 +1,46 @@
...
+    IEntityState? RawNew { get; }
+}
+
+/// <summary>
+/// Represents a state change event for a strongly typed entity and state
+/// </summary>
+/// <typeparam name="TState"></typeparam>
+/// <typeparam name="TAttributes"></typeparam>
+public interface IStateChange<TState, TAttributes> : IStateChange
+    where TAttributes : class
```
FrankBakkerNl, line 33: This seems to miss the Entity we have in the current model. It is useful to have here, but it is one of the reasons we have more type arguments there because it should also be typed correctly.

Especially when you have an
LightEntity[] lights:
lights.stateAllChanges().Subscribe(l=>l.Turn_on());

src/HassModel/NetDeamon.HassModel/Entities/IEntity.cs
Ikcelaks: The non-generic Interface is needed to support IEnumerables of entities with differing TState or TAttributes. There are definitely use cases that require that functionality.

However, I believe that I can get rid of the non-generic `EntityGeneric` class.

src/HassModel/NetDeamon.HassModel/Entities/IEntity.cs
Ikcelaks: `WithStateAs` and the version of `WithAttributesAs` that takes a custom parsing function as a parameter both work  as extension methods, but the parameterless `WithAttributesAs` method cannot be called using `entity.WithAttributesAs<TAttributesNew>()`, because the type inference fails. Writing `entity.WithAttributesAs<TAttributesNew, TState, TAttributes>()` is a real eyesore. Since that method needed to be in the IEntity interface, I put them all there. As a bonus, the implementations are slightly more compact when implemented in the class. If I could define an extension method in the interface context, I could resolve the type inference issue, but that still isn't allowed.

Here is a hypothetical use case for `WithStateAs`: You have a numeric sensor that actually only reports 5 different discrete levels despite being a unit that would appear to support continuous values. So, the (future) code generator creates a numeric entity, and you use this to view it through a custom class that interprets the values for you.

I'll code up dome examples next week.

helto4real:
What is the status of this PR? We rather not have PRs get too old. Thanks!

Ikcelaks:
Sorry! I forgot to close this.

Frank would prefer to maintain the current concrete entity classes for backwards compatibility, which is more than reasonable. However, I'm not sure how to simply accomplish that with modification of my design, so I'm dropping this PR for now and hope that it can inspire a compatible solution in the future.

PR #828, opened 2023-01-16
Add option to specify Cron Job with Seconds

spelech:
## Proposed change
Modify the CronExtensions to handle cron strings with seconds specified.  Cronos supports this but the underlying call to Parse needs to specify the format beforehand.  I opted to add a default boolean to the existing call to avoid a breaking change.

## Type of change
- [ ] Dependency upgrade

## Additional information
None

## Checklist

- [X] Local tests pass. **Your PR cannot be merged unless tests pass** -Adding addl test


src/Extensions/NetDaemon.Extensions.Scheduling/CronExtensions.cs
```diff
@@ -17,15 +17,16 @@ public static class CronExtensions
     /// <param name="scheduler">IScheduler to use for this action</param>
     /// <param name="cron">Cron expression that describes the schedule</param>
     /// <param name="action">Callback to execute</param>
+    /// <param name="hasSeconds">True if cron has Seconds specified</param>
     /// <returns>Disposable object that allows the schedule to be cancelled</returns>
-    public static IDisposable ScheduleCron(this IScheduler scheduler, string cron, Action action)
+    public static IDisposable ScheduleCron(this IScheduler scheduler, string cron, Action action, bool hasSeconds = false)
```
helto4real, line 22: Not familiar with the cron syntax on this but as a suggestion could we check for a pattern instead of providing a `hasSeconds` parameter? Even if it is not breaking it would be event better not to provide one I think.

spelech: Hmm let me give that a whirl. Cronos apparently does not try to determine if seconds are present in the string based on their function requiring the Format specifier if we use seconds.  It could be as simple as checking the number of spaces present (4-> no seconds, 5 -> has seconds).

If that has weird edge cases, I could just create a new method `ScheduleCronFine` or something to denote that is has higher resolution available

mutzl: A boolean `hasSeconds` seems absolutely fine to me.

Or alternatively directly passing the `Cronos.CronFormat`
``` C#
public static IDisposable ScheduleCron(this IScheduler scheduler, string cron, Action action, CronFormat format = CronFormat.Standard)
```
A more sophisticated parser ( = automatically detect pattern with seconds) could be a feature request in the Cronos lib.

helto4real: Agree if it complicates things a lot it is not worth it!

src/Extensions/NetDaemon.Extensions.Scheduling.Tests/Scheduling/CronTests.cs
```diff
@@ -32,8 +34,30 @@ public void TestCron()
         sub.Dispose();
         sched.AdvanceBy(TimeSpan.FromHours(1).Ticks);
         count.Should().Be(3, because: "Action should not fire after Dispose()");
+        #endregion
+
+        #region Cron with Seconds
```
FrankBakkerNl, line 39: Why not a separate test instead of regions?

Also this does not really test the seconds really work, but you could argue that is a feature or Cron.Net and not this extansion

FrankBakkerNl (changes requested):

src/Extensions/NetDaemon.Extensions.Scheduling/CronExtensions.cs
FrankBakkerNl: I Like the bool, becaue it does not expose the CRON library to the end user so we could change that in the future if needed

FrankBakkerNl (approved):

PR #852, opened 2023-03-13
MeansImplicitUse for NetDaemonAppAttribute

DerJuulsn:
## Proposed change

Adds the JetBrains MeansImplicitUse to the NetDaemonAppAttribute.
Helps R# and Rider to mark classes as used if the NetDeamonAppAttribute is used

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/src/AppModel/NetDaemon.AppModel/Common/Attributes/JetBrainsCodeAnnotations.cs
+++ b/src/AppModel/NetDaemon.AppModel/Common/Attributes/JetBrainsCodeAnnotations.cs
@@ -1,0 +1,102 @@
+/* MIT License
+
+Copyright (c) 2016 JetBrains http://www.jetbrains.com
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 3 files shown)


FrankBakkerNl:
I like this idea, If I am not correct there is a similar attributes in a MS namespace somewhere, that should probably work with both jet brains ans MS IDE and we would not have to reference JB packages

DerJuulsn:
I've added the needed code directly as an internal class as stated on this site: https://blog.jetbrains.com/dotnet/2018/05/03/what-are-jetbrains-annotations/

So we can get rid of the nuget dependency.

@FrankBakkerNl I couldn't find anything similar in the MS namespace. JetBrains normally provides info about such attributes. For example the PureAttribute: https://www.jetbrains.com/help/resharper/Reference__Code_Annotation_Attributes.html#PureAttribute

FrankBakkerNl:
This is nice indeed, now Rider will no longer show my app classes in gray :-)
Because the jetbrains code is now included as internal, I don't see a problem like this, how about you @helto4real ?

helto4real:
> This is nice indeed, now Rider will no longer show my app classes in gray :-) Because the jetbrains code is now included as internal, I don't see a problem like this, how about you @helto4real ?

Yea this is much better solution. I accept that. Good solution. Thanks!

helto4real (approved):

PR #864, opened 2023-04-24
Guide users to use WhenStateIsFor with a IScheduler instance.

### LINKED ISSUE #801: Guide users to use WhenStateIsFor with IScheduler parameter
## The problem

Currently when using WhenStateIsFor its easy to not pass in the IScheduler parameter since its a optional parameter. 

```
.WhenStateIsFor(x => x.IsOff(), TimeSpan.FromMinutes(TurnOffAfterMinutes.Value))
```

If its null then Scheduler.Default is used. This has some issues:
- Not unit testable
- When the app is disposed Scheduler.Default is not disposed so it could still call eventhandlers after the app is disposed.

## The proposed solution

Make IScheduler a non optional parameter and remove the ??. However since this is a breaking change we might want to consider offering a (obsolete) overload that does not need a IScheduler parameter. This way this change should be source compatible. Do note that this is still a binary breaking change but I think thats ok in this case?

```
.WhenStateIsFor(x => x.IsOff(), TimeSpan.FromMinutes(TurnOffAfterMinutes.Value), Scheduler)
```

Barsonax:
## Proposed change

This is a Source compatible change that will guide users to provide a ISchedule instance to WhenStateIsFor. This is however a binary breaking change but I think that might be fine as users will recompile anyway after updating netdaemon.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #801 

## Checklist

### DIFF
--- a/src/HassModel/NetDeamon.HassModel/StateObservableExtensions.cs
+++ b/src/HassModel/NetDeamon.HassModel/StateObservableExtensions.cs
@@ -37,17 +37,22 @@ public static class StateObservableExtensions
         Func<EntityState?, bool> predicate,
         TimeSpan timeSpan,
         IScheduler scheduler)
-    
-        => observable
+    {
+        if (observable == null) throw new ArgumentNullException(nameof(observable));
+        if (predicate == null) throw new ArgumentNullException(nameof(predicate));
+        if (scheduler == null) throw new ArgumentNullException(nameof(scheduler));
+        
+        return observable
             // Only process changes that start or stop matching the predicate
             .Where(e => predicate(e.Old) != predicate(e.New))
-                
+
             // Both  will restart the timer
             .Throttle(timeSpan, scheduler)
-            
+
             // But only when the new state matches the predicate we emit it
             .Where(e => predicate(e.New));
-    
+    }
+
     /// <summary>
     /// Waits for an EntityState to match a predicate for the specified time
     /// </summary>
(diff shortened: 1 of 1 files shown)


helto4real:
Not sure why you want to obsolete this? Please explain.

Barsonax:
Passing in a injected IScheduler is the more correct approach which is also unit testable.

However since ppl might be using WhenStateIsFor without IScheduler it will be breaking if you completely remove it.

helto4real:
> We discussed this some time ago when Rick created the issue for this. I agree it is preferred for several reasons to pass the Scheduler. This will now give a warning when you don't.
> 
> Someone might have a case where not passing the scheduler is 'ok'. Maybe if you only ever have short times where it is no problem if the event fires after the app unloads AND are not interested in Unit testing.
> 
> So the trade off is between preventing user from a mistake vs forcing users to pass a scheduler while it's not really needed.

I will not have user "have to" pass it. Warning should be ok. Not worth it to make it a breaking change. What you think @FrankBakkerNl ? Should we go with this?

FrankBakkerNl:
I think so, if ever we would change our minds we can just remove the obsolete attribute

src/HassModel/NetDeamon.HassModel/StateObservableExtensions.cs
```diff
@@ -7,21 +7,43 @@ namespace NetDaemon.HassModel;
...
+    
     /// <summary>
     /// Waits for an EntityState to match a predicate for the specified time
     /// </summary>
     public static IObservable<StateChange> WhenStateIsFor(
         this IObservable<StateChange> observable, 
         Func<EntityState?, bool> predicate,
         TimeSpan timeSpan,
-        IScheduler? scheduler = null)
+        IScheduler scheduler)
```
FrankBakkerNl, line 39: I think this now does require a 
ArgumentNullExcpetion.ThrowIfNull(scheduler) because you cannot rely on clients using nullable refreces

FrankBakkerNl (changes requested):

src/HassModel/NetDeamon.HassModel/StateObservableExtensions.cs
Barsonax: Yes strictly speaking that would be the case. Might as well check the other parameters as well then.

FrankBakkerNl (approved):

PR #868, opened 2023-04-29
Refactor event subscription

helto4real:
## Breaking change

This PR introduces some breaking changes. The rationale for these changes is to allow more granular subscriptions to events and remove default subscriptions to all events for users of `IHomeAssistantConnection`.  

Most users does not use the `HassClient` and should not be affected by this change.

### IHomeAssistantConnection
- `ProcessHomeAssistantEventsAsync` method no longer defaults to subscribing to all event changes. That logic is moved the new method `SubscribeToHomeAssistantEventsAsync` that returns the `IObservable<HassEvent>` instead of make it available as a property on the interface itself. 
- `OnHomeAssistantEvent` is removed and replaced by the method `SubscribeToHomeAssistantEventsAsync`
### ServiceCollectionExtensions
- `IObservable<HassEvent>` is no longer injectable through DI due to internal design change providing it through the `SubscribeToHomeAssistantEventsAsync` on the `IHomeAsstantConnection` interface instead.

## Proposed change

This PR refactors how event subscriptions work. Now it does not automatically subscribe to all events from the HomeAssistantConnection. The runtime and HassModel do add these subscriptions automatically. The function now also supports subscribing to specific avents.

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/src/Client/NetDaemon.HassClient.Tests/Integration/WebsocketIntegrationTests.cs
+++ b/src/Client/NetDaemon.HassClient.Tests/Integration/WebsocketIntegrationTests.cs
@@ -159,36 +159,167 @@ public class WebsocketIntegrationTests : IntegrationTestBase
     {
         using CancellationTokenSource tokenSource = new(TestSettings.DefaultTimeout + 1000);
         await using var ctx = await GetConnectedClientContext().ConfigureAwait(false);
-        var subscribeTask = ctx.HomeAssistantConnection
-            .OnHomeAssistantEvent
-            .Timeout(TimeSpan.FromMilliseconds(TestSettings.DefaultTimeout), Observable.Return(default(HassEvent?)))
+
+        var events = await ctx.HomeAssistantConnection
+            .SubscribeToHomeAssistantEventsAsync(null, tokenSource.Token).ConfigureAwait(false);
+
+       var subscribeTask = events
             .FirstAsync()
-            .ToTask();
+            .ToTask(tokenSource.Token);
 
         _ = ctx.HomeAssistantConnection
-            .ProcessHomeAssistantEventsAsync(tokenSource.Token);
+            .WaitForConnectionToCloseAsync(tokenSource.Token);
 
         var haEvent = await subscribeTask.ConfigureAwait(false);
 
         haEvent.Should().NotBeNull();
-        haEvent?
+        haEvent
             .EventType
             .Should()
             .BeEquivalentTo("state_changed");
 
         // We test the state changed 
-        var changedEvent = haEvent?.ToStateChangedEvent();
-        changedEvent?.EntityId
+        var changedEvent = haEvent.ToStateChangedEvent();
+        changedEvent!.EntityId
             .Should()
             .BeEquivalentTo("binary_sensor.vardagsrum_pir");
 
-        var attr = changedEvent?.NewState?.AttributesAs<AttributeTest>();
-        attr?
+        var attr = changedEvent.NewState!.AttributesAs<AttributeTest>();
+        attr!
             .BatteryLevel
             .Should()
             .Be(100);
 
-        changedEvent?.NewState?.Attributes?.FirstOrDefault(n => n.Key == "battery_level")
+        changedEvent.NewState!.Attributes!.FirstOrDefault(n => n.Key == "battery_level")!
+            .Should()
+            .NotBeNull();
+    }
+
+    [Fact]
+    public async Task TestSubscribeEventChangeAndGetEvent()
+    {
+        using CancellationTokenSource tokenSource = new(TestSettings.DefaultTimeout + 1000);
+        await using var ctx = await GetConnectedClientContext().ConfigureAwait(false);
+
+        var events = await ctx.HomeAssistantConnection
+            .SubscribeToHomeAssistantEventsAsync("state_changed", tokenSource.Token).ConfigureAwait(false);
+
+        var subscribeTask = events
+            .FirstAsync()
+            .ToTask(tokenSource.Token);
+
+        _ = ctx.HomeAssistantConnection
+            .WaitForConnectionToCloseAsync(tokenSource.Token);
+
+        var haEvent = await subscribeTask.ConfigureAwait(false);
+
+        haEvent.Should().NotBeNull();
+        haEvent
+            .EventType
+            .Should()
+            .BeEquivalentTo("state_changed");
+
+        // We test the state changed 
+        var changedEvent = haEvent.ToStateChangedEvent();
+        changedEvent!.EntityId
+            .Should()
+            .BeEquivalentTo("binary_sensor.vardagsrum_pir");
+
+        var attr = changedEvent.NewState!.AttributesAs<AttributeTest>();
+        attr!
+            .BatteryLevel
+            .Should()
+            .Be(100);
+
+        changedEvent.NewState!.Attributes!.FirstOrDefault(n => n.Key == "battery_level")!
+            .Should()
+            .NotBeNull();
+    }
+
+    [Fact]
+    public async Task TestMultipleSubscribeAllEventReturnsSameInstance()
+    {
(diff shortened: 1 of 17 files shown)


helto4real:
Still have a few tests I want to add but we can discuss the overall design first. I also have the nested task thing to take care of.

src/Client/NetDaemon.HassClient/Internal/HomeAssistantConnection.cs
```diff
@@ -49,47 +48,40 @@ IHomeAssistantApiManager apiManager
...
     }
 
-    public IObservable<HassEvent> OnHomeAssistantEvent =>
-        _hassMessageSubject.Where(n => n.Type == "event" && n.Id == _eventSubscribtionMessageId).Select(n => n.Event!);
-
-    public async Task ProcessHomeAssistantEventsAsync(CancellationToken cancelToken)
+    public async Task<IObservable<HassEvent>> SubscribeToHomeAssistantEventsAsync(string? eventType,
+        CancellationToken cancelToken)
     {
+        // When subscribe all events, optimize using the same IObservable<HassEvent> if we subscribe multiple times
```
FrankBakkerNl, line 54: Not sure if the client should be this 'smart'. If the code that uses the client wants to create multiple subscribtions for some reason that should also be possible.

src/Client/NetDaemon.HassClient/Internal/HomeAssistantConnection.cs
```diff
@@ -49,47 +48,40 @@ IHomeAssistantApiManager apiManager
...
+        return observableResult;
+    }
 
-            return resultEvent;
-        }
-        finally
-        {
-            _messageIdSemaphore.Release();
-        }
+    public async Task ProcessHomeAssistantEventsAsync(CancellationToken cancelToken)
```
FrankBakkerNl, line 76: This feels a bit strange now. It is basically only awaiting the internal _internalCancelSource or the provided one and nopt doing any actual work itself. _internalCancelSource seems to only get cancelled when the connection is Disposed

src/Client/NetDaemon.HassClient.Tests/Integration/WebsocketIntegrationTests.cs
```diff
@@ -193,13 +238,104 @@ public async Task TestSubscribeAndGetEvent()
...
+            .EventType
+            .Should()
+            .BeEquivalentTo("state_changed");
+        secondHaEvent?
+            .EventType
+            .Should()
+            .BeEquivalentTo("state_changed");
+
+        // We test the state changed 
+        var firstChangedEvent = firstHaEvent?.ToStateChangedEvent();
```
FrankBakkerNl, line 298: Use ! iso ? in asserts, otherwise the Should() is not executed when there is a null

src/Client/NetDaemon.HassClient.Tests/Integration/WebsocketIntegrationTests.cs
```diff
@@ -193,13 +238,104 @@ public async Task TestSubscribeAndGetEvent()
...
+        var secondEventSubscriptionTask = secondEventSubscription
+            .Timeout(TimeSpan.FromMilliseconds(TestSettings.DefaultTimeout), Observable.Return(default(HassEvent?)))
+            .FirstAsync()
+            .ToTask();
+
+        // Since we are not subscribing to all events we should have two different instances
+        Assert.NotSame(firstEventSubscription, secondEventSubscription);
+
+        _ = ctx.HomeAssistantConnection
+            .ProcessHomeAssistantEventsAsync(tokenSource.Token);
```
FrankBakkerNl, line 280: I think these calls are not really needed anymore and maybe even the entire ProcessHomeAssistantEventsAsync method can be removed

src/HassModel/NetDaemon.HassModel.Tests/Internal/EntityAreaCachTests.cs
```diff
@@ -21,6 +21,11 @@ public async Task EntityIdWithArea_Returns_HassArea()
 
         haRunnerMock.SetupGet(n => n.CurrentConnection).Returns(_hassConnectionMock.Object);
 
+        _hassConnectionMock.Setup(n =>
+                n.SubscribeToHomeAssistantEventsAsync(It.IsAny<string>(), It.IsAny<CancellationToken>()))
+            .ReturnsAsync(testSubject
+            );
```
FrankBakkerNl, line 27: This ')' closes te one at line 26, so it makes more sense to remove the newline (same for all below and other files) (but maybe this is not even needed anymore If yoy could remove the ProcessHomeAssistantEventsAsync() method)

src/HassModel/NetDeamon.HassModel/Internal/EntityAreaCache.cs
```diff
@@ -27,7 +27,7 @@ public async Task InitializeAsync(CancellationToken cancellationToken)
     {
         _cancellationToken = cancellationToken;
 
-        var events = _provider.GetRequiredService<IObservable<HassEvent>>();
+        var events = await _hassRunner.CurrentConnection!.SubscribeToHomeAssistantEventsAsync(null, _cancellationToken).ConfigureAwait(false);
```
FrankBakkerNl, line 30: Ah, so this relies on the fact that the connection caches the subscription for all events. I think it is better to have the Runner provide a method or property to get the Observable<HassEvent> that is cached there

src/HassModel/NetDeamon.HassModel/Internal/EntityStateCache.cs
```diff
@@ -37,7 +37,7 @@ public async Task InitializeAsync(CancellationToken cancellationToken)
     {
         _ = _hassRunner.CurrentConnection ?? throw new InvalidOperationException();
 
-        var events = _provider.GetRequiredService<IObservable<HassEvent>>();
+        var events = await _hassRunner.CurrentConnection!.SubscribeToHomeAssistantEventsAsync(null,  cancellationToken).ConfigureAwait(false);
```
FrankBakkerNl, line 40: Same here as for EntityAreaCache

src/Client/NetDaemon.HassClient/Internal/HomeAssistantConnection.cs
helto4real: I do want this optimization here for Client users but you can provide a "*" and get all events multiple times if you want to. Move this from Client will take a lot of more work. We could improve this further in future PR:s. I did change it to do lazy loading and thread safe.

src/Client/NetDaemon.HassClient.Tests/Integration/WebsocketIntegrationTests.cs
helto4real: Good find!

src/HassModel/NetDaemon.HassModel.Tests/Internal/EntityAreaCachTests.cs
helto4real: Fixed this. I think it was some weird autoformatting thing

src/HassModel/NetDeamon.HassModel/Internal/EntityAreaCache.cs
helto4real: I decided to keep the cached thing. See that comment.

src/HassModel/NetDeamon.HassModel/Internal/EntityStateCache.cs
helto4real: See comments before.

src/Client/NetDaemon.HassClient.Tests/Integration/WebsocketIntegrationTests.cs
helto4real: See other comment

helto4real:
I missed a few spots in the tests and also fixed old tests that had the same problem. I will merge this now. Thanks for the review @FrankBakkerNl

Issue #896, opened 2023-07-22
Entity interfaces to use generated code in external library


x00Pavel:
## The problem
There is no way how to use/reference generated code in an external library shipped as a Nuget package. 
If I want to make some automation (library) based on generated code (especially, data types and extension methods), this automation has to be shared as source code from GitHub as the Nuget package won't be packed as this external library doesn't know anything about generated code.

## The proposed solution
In generated code use interfaces that are defined in the HASS model.
Then use defined interfaces in the external library.

## The alternatives
Move common extension methods to the HASS model and refactor them to use mentioned Interfaces.

## Additional context
The biggest problem is in extension methods as there is no way how to define in interfaces possible extension methods.

FrankBakkerNl:
Solved with the CoreInterfaces feature

PR #897, opened 2023-07-22
Add interfaces

x00Pavel:
## Proposed change

Add base interfaces for generated Entities.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: #896 

This functionality is discussed in Discord thread `Entity interfaces in NetDaemon`

## Checklist


src/HassModel/NetDeamon.HassModel/Entities/Interfaces/IOnOffTarget.cs
```diff
@@ -0,0 +1,139 @@
+namespace NetDaemon.HassModel.Entities;
+
+#pragma warning disable CA1040
+
+/// <summary>
+/// 
+/// </summary>
+public interface IToggleParameters
```
FrankBakkerNl, line 8: I don't think these interfaces are needed

src/HassModel/NetDeamon.HassModel/Entities/Interfaces/IOnOffTarget.cs
```diff
@@ -0,0 +1,139 @@
...
+public interface ITurnOnParameters
+{
+}
+
+
+
+/// <summary>
+/// Base interface for togglebal targets
+/// </summary>
+public interface IOnOffTarget
```
FrankBakkerNl, line 31: This could be derived from IServiceTarget. That would even make it possible to have a default implementation of these methods that actually works

src/HassModel/NetDeamon.HassModel/Entities/Interfaces/IOnOffTarget.cs
```diff
@@ -0,0 +1,139 @@
...
+    ///<param name="colorTemp">Color temperature for the light in mireds.</param>
+    ///<param name="kelvin">Color temperature for the light in Kelvin.</param>
+    ///<param name="brightness">Number indicating brightness, where 0 turns the light off, 1 is the minimum brightness and 255 is the maximum brightness supported by the light.</param>
+    ///<param name="brightnessPct">Number indicating percentage of full brightness, where 0 turns the light off, 1 is the minimum brightness and 100 is the maximum brightness supported by the light.</param>
+    ///<param name="white">Set the light to white mode.</param>
+    ///<param name="profile">Name of a light profile to use. eg: relax</param>
+    ///<param name="flash">If the light should flash.</param>
+    ///<param name="effect">Light effect.</param>
+    public static void Toggle(long? transition = null, object? rgbColor = null, object? colorName = null, object? hsColor = null, object? xyColor = null, object? colorTemp = null, long? kelvin = null, long? brightness = null, long? brightnessPct = null, object? white = null, string? profile = null, object? flash = null, string? effect = null)
```
FrankBakkerNl, line 92: The TurnOn and Toggle with all these arguments are specific to lights, not for all OnOff Entities

src/HassModel/NetDeamon.HassModel/Entities/Interfaces/IServiceTarget.cs
```diff
@@ -0,0 +1,14 @@
+namespace NetDaemon.HassModel.Entities;
+
+/// <summary>
+/// 
+/// </summary>
+public interface IServiceTarget: IEntity
```
FrankBakkerNl, line 6: Should not derive from IEntity. I don't think we need IEntity for this, I do want to introduce it later maybe as part of another feature so preferably I would not add it now and lock into a specific design yet

FrankBakkerNl (changes requested):

src/HassModel/NetDeamon.HassModel/Entities/Interfaces/IOnOffTarget.cs
x00Pavel: I added them to replace generated parameter classes. Without these, I wouldn't be able to specify methods with Parameter class.

src/HassModel/NetDeamon.HassModel/Entities/Interfaces/IServiceTarget.cs
x00Pavel: Then `IServiceTarget` would need to have some attributes from `IEntity`. This is because of `EnumerableEntityExtensions` class. Inside it, there is a usage of `HaContext` and `StateAllChanges()` methods of the Entity class. This is the reason why I added this derivation. If you have any suggestion how to solve it, I'm don't have anything against it.

src/HassModel/NetDeamon.HassModel/Entities/Interfaces/IOnOffTarget.cs
FrankBakkerNl: The methods with parameter types will be in the generated code, so they are not needed here.

x00Pavel:
I've removed all methods from the interface to simplify the PR. To have empty interfaces is enough to fulfill needs of my request.

x00Pavel:
btw should I squash all commits?

FrankBakkerNl:
You can, but eventually we will squash commit the whole PR anyway. Keeping then separate at first makes reviewing changes easier

FrankBakkerNl (approved):

src/HassModel/NetDaemon.HassModel.CodeGenerator/NetDaemon.HassModel.CodeGenerator.csproj
```diff
@@ -18,7 +18,6 @@
         </PackageDescription>
         <PackageReleaseNotes>Please advice this is still in beta</PackageReleaseNotes>
         <tags>Home Assistant</tags>
-		<GenerateDocumentationFile>True</GenerateDocumentationFile>
```
helto4real, line 21: Why is this removed? We added docs just recently. Or am I missing something?

src/HassModel/NetDaemon.HassModel.CodeGenerator/CodeGeneration/EntitiesGenerator.cs
```diff
@@ -18,15 +18,12 @@ public static IEnumerable<MemberDeclarationSyntax> Generate(IReadOnlyCollection<
...
         }
-        
         foreach (var domainMetadata in metaData)
         {
             yield return GenerateEntityType(domainMetadata);
-            
             yield return AttributeTypeGenerator.GenerateAttributeRecord(domainMetadata);
         }
     }
-
```
helto4real, line 29: Ther should be one line between functions.

src/debug/MyNDApp/apps/HassModel/MyInterfaceAutomation/InterfaceUsage.cs
```diff
@@ -0,0 +1,28 @@
...
+using Microsoft.Extensions.Logging;
+using MyLibrary;
+using NetDaemon.AppModel;
+using NetDaemon.HassModel;
+using NetDaemon.HassModel.Entities;
+
+namespace Debug.apps.HassModel.MyInterfaceAutomation;
+
+[NetDaemonApp]
+[Focus]
```
helto4real, line 14: Please remove Focus attribute before merge

helto4real (approved):
Just some small stuff to fix before merge if you want to. I approve and let me know when it is ready for merge.

src/HassModel/NetDaemon.HassModel.CodeGenerator/NetDaemon.HassModel.CodeGenerator.csproj
FrankBakkerNl: This is the codegen, it does not need API docs as we do not provide it as a nuget to be used in usercode

helto4real:
I got the answer from Frank regarding the docs removal. Rest is not important to fix. Merging

PR #898, opened 2023-07-31
Cleaned some extensions

svrooij:
## Proposed change

- Support for User Secrets by not overriding it.

## Type of change

- [ ] Dependency upgrade

## Additional information

As discussed on discourt: https://discord.com/channels/686787269780176998/690899481046024233/1135518233458393108

## Checklist

### DIFF
--- a/src/Runtime/NetDaemon.Runtime/Common/Extensions/HostBuilderExtensions.cs
+++ b/src/Runtime/NetDaemon.Runtime/Common/Extensions/HostBuilderExtensions.cs
@@ -7,29 +7,53 @@ namespace NetDaemon.Runtime;
 
 public static class HostBuilderExtensions
 {
+    /// <summary>
+    /// Call this method to load NetDeamonYaml settings, and to register 'ConfigureNetDaemonServices' in the service collection
+    /// </summary>
+    /// <param name="hostBuilder"></param>
+    /// <remarks>
+    /// UseNetDaemonAppSettings has several responsibilities:
+    ///  - Register appsettings.json to the host configuration
+    ///  - Register all the yaml settings from the path set in the current configuration to the configuration provider
+    ///  - Call 'ConfigureNetDaemonServices' in the service collection
+    ///  
+    /// You can call these methods separately if you want to do something else in between, or if you're calling any of these methods already.
+    /// Change `UseNetDaemonAppSettings` to `.RegisterAppSettingsJsonToHost().RegisterYamlSettings()` and call `ConfigureNetDaemonServices(context.Configuration)` in ConfigureServices.
+    /// </remarks>
     public static IHostBuilder UseNetDaemonAppSettings(this IHostBuilder hostBuilder)
     {
         return hostBuilder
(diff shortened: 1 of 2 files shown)


svrooij:
After some thought I think the original code already supported user secrets, they were just overridden. The last config provider takes precedence. the `CreateDefaultBuilder` sets:

- Environment
- commandline arguments
- appsettings.json and appsettings.{environment}.json
- User secrets (if set to development)
- Environment variables
- commandline arguments

Then the `UseNetDaemonAppSettings` adds the following (which thus override the previous set):
- appsettings.json and appsettings.{env}.json
- environment
- yaml

This means that if you would remove the `token` from the app settings it would not get overriden. Either way, no need to set the configuration providers twise.

src/Runtime/NetDaemon.Runtime/Common/Extensions/HostBuilderExtensions.cs
```diff
@@ -7,29 +7,53 @@ namespace NetDaemon.Runtime;
...
             .ConfigureServices((context, services)
                 => services.ConfigureNetDaemonServices(context.Configuration)
-            )
-            .ConfigureAppConfiguration((ctx, config) =>
-            {
-                // TODO: Most of this seems to be what Host.CreateDefaultBuilder already does 
-                config.SetBasePath(Directory.GetCurrentDirectory());
-                config.AddJsonFile("appsettings.json");
-                config.AddJsonFile($"appsettings.{ctx.HostingEnvironment.EnvironmentName}.json", true);
-                config.AddEnvironmentVariables();
```
helto4real, line 22: This is missing? It is important that we still support using environment vars since the default container use them

helto4real (changes requested):
Looks good to me but I am not sure if environment is still working without the `AddEnvironmentVariables()` can you make sure that is not broken?

src/Runtime/NetDaemon.Runtime/Common/Extensions/HostBuilderExtensions.cs
svrooij: I’m 100% sure they still work.

source of the CreateDefaultBuilder: https://github.com/dotnet/runtime/blob/e43cba31e997e960205844b3f0541011ef732c29/src/libraries/Microsoft.Extensions.Hosting/src/HostingHostBuilderExtensions.cs#L256

But I also don’t trust the source, so I set a breakpoint inside the new RegisterYaml… and inspected the available config providers.

svrooij:
The failing test is strange, they all succeed locally, I explicitly checked.

svrooij:
@helto4real do you know how it’s possible for the coverage to go down if I removed code? That means the old code was not tested I guess? Want me to add tests for the yams settings part?

helto4real:
> The failing test is strange, they all succeed locally, I explicitly checked.

No it is ok! Sometimes it does this. The code was never tested in the first place and I am not sure how you would test this efficiently.

helto4real (approved):
Ok thanks for clarification and the PR!

PR #944, opened 2023-09-28
Move all entity methods to the IEntityCore interface

### LINKED ISSUE #943: `Entity` methods are no accessable in external libraries via interfaces
## The problem

`StateChanges` method of `Entity` class is not available for usage with child interfaces for `IEntityCore`.

## The proposed solution

Add all general methods from `Entity` class to `IEntityCore` interface

x00Pavel:
## Proposed change

All methods and class attributes are moved to `IEntityCore` interface to have ability for external libraries access these methods and attributes

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #943 

## Checklist

### DIFF
--- a/src/HassModel/NetDeamon.HassModel/Entities/Core/IEntityCore.cs
+++ b/src/HassModel/NetDeamon.HassModel/Entities/Core/IEntityCore.cs
@@ -1,7 +1,7 @@
 namespace NetDaemon.HassModel.Entities;
 
 /// <summary>
-/// Core  interface for any entity
+/// Core interface for any entity
 /// </summary>
 public interface IEntityCore
 {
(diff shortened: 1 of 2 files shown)


FrankBakkerNl (changes requested):
This was intended

Rora:
@FrankBakkerNl I was looking to propose a similar change (using more interfaces so that developers can mock away parts depending on their needs) and saw this PR.

I see that this PR was closed, I can't find the rationale of not adding methods to the interface so I don't know if it would make sense to propose a PR to add other methods/properties. Could you help me understand the reasoning behind this?

I think it would make sense to have generated classes like `InputBooleanEntities` implement interfaces and have them reference the entities via interfaces as well, would love to hear your thoughts on this.

FrankBakkerNl:
The reason we have the IEntityCore and derived interfaces is not for testing, but to enable interaction between generated code and 3rd party libraries. The generated entity classes implement these interfaces as a 'marker interface'. A library can then eg expose a method that takes in an an ICalendarEntityCore to allow it to be called with a generated entity instance. Internally the library could then convert it to its own generated CalendatEntity type. Therefore we keep this interface to the minimun needed (core) to be able to us in a copy contructor, that is just the EntityId and the IHaContext.

A great way to write tests is to mock IHaContext instead of mocking individual entities. An entity is actually just a simple wrapper on top of IHaContext. Mocking IHaContext will allow apps to be tested to also use the generated navigation properties (Enties.Lights.LivingRoom.GetState() etc.). Look at some examples in https://github.com/net-daemon/awesome_netdaemon and https://github.com/FrankBakkerNl/NetDaemonExample

PR #954, opened 2023-10-17
IEntityCore exposes all Entity functionality, EntityExtensions operate on interface

zeroflag0:
### Why?
Writing (additional) Extensions for (generated) `Entity`s is troublesome.
All methods in `EntityExtension` operate on `Entity`, but not with interfaces.
However generated code operates on interfaces, but `IEntityCore` exposes virtually nothing of the common base-class `Entity`.

### Simple Solution
Adding all of `Entity`'s interface to `IEntityCore`, making the functionallity universally assessable.
Changing `EntityExtensions` to operate on `IEntityCore`, making it compatible with generated-entity-inheritance.

### Why it shouldn't break anything
The only way this could break anything, is when a generated `Entity` would NOT use `Entity` as a `base` class - hence missing the functionality required by the changed `IEntityCore`.
From what I can see, all generated entities use `Entity` as `base` class, so there's no reason this could be breaking anything. (also works with anything that was generated for me)

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

- [x] _Tests have been added to verify that the new code works._ -> `this is an API-enhancement, unit-testing does not apply`

- [x] _Documentation added/updated for [www.home-assistant.io][docs-repository]_ -> `not relevant because API is documented itself`


FrankBakkerNl:
These members were deliberately not added here to avoid possible future breakiung changes. What exactly is the scenario you want to support here?

helto4real:
@zeroflag0 thank you for your PR, please if you want this PR to get merged please answer the question from @FrankBakkerNl . After you would have to rebase it to match latest .NET 8 version that now are currently in dev branch. Thanks

FrankBakkerNl:
To be clear on expectations, I do not think we will merge this. We had a similar PR before which we also did not merge. I am interested in the exact scenario you would like supported, so we can look at how we could support that.

zeroflag0:
Hi @FrankBakkerNl , @helto4real ,
sorry for taking this long to answer. I was trying to recreate a use case, where Generics-Magic (with Entity<TEntity,EntityState<...>,...>) would not suffice, and the only thing I found was the NumericEntityState - but then my extension would not fix that either.

Thinking about it: Having interface would make it a lot easier, to write extensions - but are not strictly necessary.
I prefer using interfaces wherever possible (in single inheritance languages like C# anyway) because it allows the API-user to implement a class with any base-class, that still provides all interfaces required for extension methods and the likes.

I guess such a (minor) "design change" would require a little more than what my PR does, but maybe I can provide some more interfaces in the next week(s).
General idea would be to have IEntityState<TAttributes, TState>, IEntity<TEntityState> and stuff like that, and then provide explicit interface-implementation for <object?,string?> for the Entity(non-generic) and EntityState(non-generic) variant/base class.
What do you think? (I'd provide a first draft for that)

FrankBakkerNl:
The current genetics on Entity, EntityState and EntityStateChangs are pretty complicated. This has in part to do with the fact that the API is circulair. Entity has IObservable<StateChange> which then also needs a TEntity argument. That entity needs to be of a specific type as well so you can use it typed in the state change eventhandler.

I have played with IEntity<out TAttributes, ..> and other interfaces to allow a more flexible model. Interfaces allow covariance, where classes do not which is an advantage here. Mainly because all kind of helper methods could be simplified.

Making a new good API and doing this in a backwards compatible way is a challenge. I would consider breaking changes if we also have a lot of benefits and are really happy with that new design. At this point I am not there yet

FrankBakkerNl:
You are ofc welcome to draft something up. I got kind of stuck thinking about this, so new insights are welcome. Are you on our Discord server? That would be the best place to discuss details

zeroflag0:
I am on the ND discord - though I don't launch the app that often.
Handle is tamira145

I'll get in touch once I have the time to draft something - currently stuck on IRL issues though. 😉

FrankBakkerNl:
Great, will close this now. Whenever you have time we can pick this up

PR #1054, opened 2024-02-18
Fixed clean stopping when connection error halts the ending of runtime

helto4real:
## Proposed change

This was an annoying bug that made the runtime wait infinite time if there was an error in the connection phase. 

The bug could potentially lead to deadlock situation where the service could not be stopped if there was a connection problem too. 

Something changed the timings with this change so I had to make a small delay before the integration tests will run or the state cache will not be populated and tests would fail. I really have no idea how to make it more robust without using delays. It is fixed for now but if the tests will be flaky I will have to look at a long term solution.

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/src/Runtime/NetDaemon.Runtime/Internal/NetDaemonRuntime.cs
+++ b/src/Runtime/NetDaemon.Runtime/Internal/NetDaemonRuntime.cs
@@ -57,7 +57,7 @@ internal class NetDaemonRuntime(IHomeAssistantRunner homeAssistantRunner,
                 _runnerCancellationSource.Token);
 
             // Make sure we only return after the connection is made and initialization is ready
-            await _startedAndConnected.Task;
+            await Task.WhenAny( _startedAndConnected.Task, _runnerTask);
         }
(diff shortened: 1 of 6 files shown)


src/Runtime/NetDaemon.Runtime/Internal/NetDaemonRuntime.cs
```diff
@@ -80,12 +81,10 @@ public async Task StartAsync(CancellationToken stoppingToken)
                 return;
             }
 
-            IsConnected = true;
-
             await cacheManager.InitializeAsync(cancelToken).ConfigureAwait(false);
 
             await LoadNewAppContextAsync(haConnection, cancelToken);
-
+            IsConnected = true;
```
FrankBakkerNl, line 87: I remember making changes here some time ago (maybe those caused the issue you had now) 
I think I remember It was imporrtant rthat IsConnacted was set to true before the apps are loaded so that they can use the Connection in their startup code

PR #1093, opened 2024-05-07
Add data to service extensions for services that lacks metadata

helto4real:
## Proposed change

By adding the possibility to add custom data for services generated as extension methods on entities we can support service calls for services that does not provide metadata.

Example:
Calling a script can be called with custom data but since it lacks metadata the codegenerator generates empty calls.

Before:
```csharp
public static class ScriptEntityExtensionMethods
{
...

    ///<summary>Runs the sequence of actions defined in a script.</summary>
    public static void TurnOn(this IScriptEntityCore target)
    {
        target.CallService("turn_on");
    }
...
}
```
After:
```csharp
public static class ScriptEntityExtensionMethods
{
...

    ///<summary>Runs the sequence of actions defined in a script.</summary>
    public static void TurnOn(this IScriptEntityCore target, object? data = null)
    {
        target.CallService("turn_on", data);
    }
...
}
```

Also the same changes to servicecall. We add a data to service calls with no metadata but keeps the one  with metadata the same:

Example of new generated optional data provided.
```csharp
public partial class ScriptServices
{
    private readonly IHaContext _haContext;
    public ScriptServices(IHaContext haContext)
    {
        _haContext = haContext;
    }

    ///<summary>Reloads all the available scripts.</summary>
    public void Reload(object? data = null)
    {
        _haContext.CallService("script", "reload", null, data);
    }

    public void Setnightmode(object? data = null)
    {
        _haContext.CallService("script", "setnightmode", null, data);
    }

    ///<summary>Toggle a script. Starts it, if isn&apos;t running, stops it otherwise.</summary>
    ///<param name="target">The target for this service call</param>
    public void Toggle(ServiceTarget target, object? data = null)
    {
        _haContext.CallService("script", "toggle", target, data);
    }

    ///<summary>Stops a running script.</summary>
    ///<param name="target">The target for this service call</param>
    public void TurnOff(ServiceTarget target, object? data = null)
    {
        _haContext.CallService("script", "turn_off", target, data);
    }

    ///<summary>Runs the sequence of actions defined in a script.</summary>
    ///<param name="target">The target for this service call</param>
    public void TurnOn(ServiceTarget target, object? data = null)
    {
        _haContext.CallService("script", "turn_on", target, data);
    }
}
```

The code should not be breaking because of the optional data parameter.

## Type of change

- [ ] Dependency upgrade
- [ ] Breaking change (fix/feature causing existing functionality to break)c

## Additional information

## Checklist

### DIFF
--- a/src/HassModel/NetDaemon.HassModel.CodeGenerator/CodeGeneration/ServicesGenerator.cs
+++ b/src/HassModel/NetDaemon.HassModel.CodeGenerator/CodeGeneration/ServicesGenerator.cs
@@ -97,7 +97,14 @@ internal static class ServicesGenerator
 
         if (serviceArguments is null)
         {
-            targetParam = "object? data";
+            if (service.Target is not null)
+            {
+                targetParam = $"{targetParam}, object? data = null";
+            }
+            else
+            {
+                targetParam = "object? data = null";
+            }
             targetArg = "null, data";
             // method without arguments
             yield return ParseMemberDeclaration($$"""
--- a/src/HassModel/NetDaemon.HassModel.CodeGenerator/CodeGeneration/ServicesGenerator.cs
+++ b/src/HassModel/NetDaemon.HassModel.CodeGenerator/CodeGeneration/ServicesGenerator.cs
@@ -97,6 +97,8 @@ internal static class ServicesGenerator
 
         if (serviceArguments is null)
         {
+            targetParam = "object? data";
+            targetArg = "null, data";
             // method without arguments
             yield return ParseMemberDeclaration($$"""
                         void {{serviceMethodName}}({{targetParam}})
--- a/src/HassModel/NetDaemon.HassModel.CodeGenerator/CodeGeneration/ServicesGenerator.cs
+++ b/src/HassModel/NetDaemon.HassModel.CodeGenerator/CodeGeneration/ServicesGenerator.cs
@@ -97,21 +97,11 @@ internal static class ServicesGenerator
 
         if (serviceArguments is null)
         {
-            if (service.Target is not null)
-            {
-                targetParam = $"{targetParam}, object? data = null";
-                targetArg = "target, data";
-            }
-            else
-            {
-                targetParam = "object? data = null";
-                targetArg = "null, data";
-            }
             // method without arguments
             yield return ParseMemberDeclaration($$"""
-                        void {{serviceMethodName}}({{targetParam}})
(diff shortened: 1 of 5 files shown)


FrankBakkerNl (approved):

FrankBakkerNl:
Just after approving I thought that it would need some basic tests

helto4real:
> Looks easy enough, just needs one or two tests to check the code compiles with and without pasing an argument here

I agree I will do tests :)

src/HassModel/NetDaemon.HassModel.CodeGenerator/CodeGeneration/ServicesGenerator.cs
```diff
@@ -97,6 +97,8 @@ private static IEnumerable<MemberDeclarationSyntax> GenerateServiceMethod(string
 
         if (serviceArguments is null)
         {
+            targetParam = "object? data";
```
FrankBakkerNl, line 100: I think traget is the ServiceTarget, eg the Entitiy, Area, Floor etc. This can be present or not regardless of how many servcieArguments there are. So you need to join them like is done at the third yield. (maybe reorganize so the code does not need to be duplicated)

helto4real: This is the case when there are no service arguments, today it does not generate any input. Not sure I follow. I will check it out.

helto4real:
```csharp
 // method using arguments as separate parameters
            yield return ParseMemberDeclaration($$"""
                        void {{serviceMethodName}}({{JoinList(targetParam, serviceArguments.GetParametersList())}})
```
that third yield is when there are service targets. This is the not the case here. I only want to add the data option if there are no targets.

FrankBakkerNl: It is the correct case indeed, just that you now override the serviceTarget with the data parameter.

helto4real: I added tests and the case you were talking about in Discord. Actually found the problem when I did the test 👍

helto4real:
Merged after Frank gave ok on discord

PR #1094, opened 2024-05-10
Deprecates the SetEntityState in the HassModel.Integration assembly

helto4real:
## Proposed change

We have a preferred way to create entities using the MQTT extension. The integration used internal APIs and are bound to break at some point. Also we should not have multiple ways to do things that we have to maintain. 

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/src/HassModel/NetDaemon.HassModel.Tests/.editorconfig
+++ b/src/HassModel/NetDaemon.HassModel.Tests/.editorconfig
@@ -2,4 +2,4 @@
 dotnet_diagnostic.xUnit1030.severity = none
 # Warning CA1861 : Prefer 'static readonly' fields over constant array arguments if the called method is called repeatedly and is not mutating the passed array (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1861)
 dotnet_diagnostic.CA1861.severity= none
-
+dotnet_diagnostic.CS0618.severity = none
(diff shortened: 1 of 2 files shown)


src/HassModel/NetDaemon.HassModel.Tests/.editorconfig
```diff
@@ -2,4 +2,4 @@
 dotnet_diagnostic.xUnit1030.severity = none
 # Warning CA1861 : Prefer 'static readonly' fields over constant array arguments if the called method is called repeatedly and is not mutating the passed array (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1861)
 dotnet_diagnostic.CA1861.severity= none
-
+dotnet_diagnostic.CS0618.severity = none
```
FrankBakkerNl, line 5: Maybe put this just in the code of the test where we need it, so it will also be removed when we eventually remove the method?

Barsonax: I think disabling obsolete warnings in tests makes sense though. Even if you mark something obsolete you still want to test it right?

Then when you remove it it will be a compile error anyway.

FrankBakkerNl: Agree to suppress the warning because obviously you still want to test it. I was just thinking to suppress in code just the specific obsolete methods in our own code we want to test iso suppressing all obsolete messages. If for instance now Fluent Assertions would make something obsolete we would not know

Barsonax: That's indeed a good case for not ignoring it completely. Though it probably won't happen that often.

Would be nice if there was some way to make a distinction between obsolete in our own code and from another library.

helto4real: I was more thinking like @Barsonax. We never want those warnings in tests. I tried to do per-test exclusion but somehow it did not work @FrankBakkerNl

FrankBakkerNl (approved):

PR #1129, opened 2024-06-30
Make restart logic progressive and other small fixes

helto4real:
## Proposed change

Fixes in this PR:
### Progressive reconnect timeouts
To avoid too many tries it will increase the timeout per retry by a factor 2 starting with 5 seconds and max 80 seconds.

### The registry updated log will now be a debug log
Recent versions of HA updates the registries more often and possible flooding the outputlog. This log is now made a debug log instead.

### Fix the problem that the correct version does not show on start
The github action that push to Nuget did not successfully updated version information. The default message also changed  to reflect the actual behaviour. The version is not known when using a local built NetDaemon runtime. The version will then be `local build`. Offical built versions will show the correct version.

### Small fixes in restart logic
- The runtime stops when NetDaemon not able to authorize the token. This is same behavior as before but now the runtime shuts down too.
- Do not show timeout information on disconnect since this is also shown in the reconnect loop
 
## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/src/Client/NetDaemon.HassClient/Internal/Helpers/ProgressiveTimeout.cs
+++ b/src/Client/NetDaemon.HassClient/Internal/Helpers/ProgressiveTimeout.cs
@@ -1,7 +1,7 @@
 namespace NetDaemon.Client.Internal.Helpers;
 
 /// <summary>
-///  Progeressively increases timeout from a min valuye to a max value with a step
+///  Progeressively increases timeout from a min value to a max value with a step
 /// </summary>
 public class ProgressiveTimeout
 {
--- a/src/Client/NetDaemon.HassClient/Internal/Helpers/ProgressiveTimeout.cs
+++ b/src/Client/NetDaemon.HassClient/Internal/Helpers/ProgressiveTimeout.cs
@@ -1,0 +1,41 @@
+namespace NetDaemon.Client.Internal.Helpers;
+
+/// <summary>
+///  Progeressively increases timeout from a min valuye to a max value with a step
+/// </summary>
+public class ProgressiveTimeout
+{
+    private readonly int _initialTimeout;
+    private readonly int _maxTimeout;
+    private readonly double _increaseFactor;
+    private int _currentTimeout;
+
+    public ProgressiveTimeout(int initialTimeout, int maxTimeout, double increaseFactor)
+    {
+        if (initialTimeout <= 0)
+            throw new ArgumentOutOfRangeException(nameof(initialTimeout), "Initial timeout must be greater than zero.");
+
+        if (maxTimeout < initialTimeout)
+            throw new ArgumentOutOfRangeException(nameof(maxTimeout), "Max timeout must be greater than or equal to initial timeout.");
+
+        if (increaseFactor <= 1)
+            throw new ArgumentOutOfRangeException(nameof(increaseFactor), "Increase factor must be greater than 1.");
+
+        _initialTimeout = initialTimeout;
+        _maxTimeout = maxTimeout;
+        _increaseFactor = increaseFactor;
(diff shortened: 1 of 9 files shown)


helto4real:
The CI is failing due to a deprecated call that been removed in the beta. Another pr fixes this.

src/Client/NetDaemon.HassClient/Internal/HomeAssistantRunner.cs
```diff
@@ -60,13 +62,17 @@ public async ValueTask DisposeAsync()
     private async Task InternalRunAsync(string host, int port, bool ssl, string token, string websocketPath, TimeSpan timeout,
         CancellationToken cancelToken)
     {
+        var progressiveTimeout = new ProgressiveTimeout((int)timeout.TotalSeconds, MaxTimeoutInSeconds, 2.0);
```
FrankBakkerNl, line 65: The time span comes in as a parameter, which implies the caller of this method can use any timespan he wants, not just a round number of seconds. Allowing this as a parameter, by then not using that specific value but the progressive Timeout seems like not doing what the caller expects

FrankBakkerNl: ProgressiveTimeout could be passed a Timespan iso an int, or maybe the ticks iso the seconds to avoid unneccary rounding

helto4real: Good suggestions, I fixed that now.

src/Client/NetDaemon.HassClient/Internal/Helpers/ProgressiveTimeout.cs
```diff
@@ -0,0 +1,45 @@
...
+public class ProgressiveTimeout
+{
+    private readonly TimeSpan _initialTimeout;
+    private readonly TimeSpan _maxTimeout;
+    private readonly double _increaseFactor;
+    private TimeSpan _currentTimeout;
+
+    public ProgressiveTimeout(TimeSpan initialTimeout, TimeSpan maxTimeout, double increaseFactor)
+    {
+        if (initialTimeout <= TimeSpan.Zero)
```
Barsonax, line 15: I would use ArgumentOutOfRangeException.ThrowIfLessThanOrEqual here for 2 reasons:
- Its simpler
- Leads to faster code due to the method size being smaller (throwing an exception makes the method body bigger)

src/Client/NetDaemon.HassClient/Internal/Helpers/ProgressiveTimeout.cs
```diff
@@ -0,0 +1,45 @@
...
+            throw new ArgumentOutOfRangeException(nameof(initialTimeout), "Initial timeout must be greater than zero.");
+
+        if (maxTimeout < initialTimeout)
```
Barsonax, line 18: I think the comparison here should be less than or equal so it matches the message

Barsonax (changes requested):

src/Client/NetDaemon.HassClient/Internal/Helpers/ProgressiveTimeout.cs
helto4real: Yea I tend forget these exists... thanks!

src/Client/NetDaemon.HassClient/Internal/Helpers/ProgressiveTimeout.cs
helto4real: Good find this is  a logic error indeed.

Barsonax (approved):

PR #1131, opened 2024-07-04
Stability fixes

helto4real:
## Proposed change

Some quality and stability fixes.
- [x] Make the dispose of application context thread safe
- [x] Correct sample of use calendar in debug project
- [x] Make the handling of usage of cancellation token that could potentially be used from a disposed or canceled Cancellation token source, check screenshot

image

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist


src/AppModel/NetDaemon.AppModel/Internal/Context/ApplicationContext.cs
```diff
@@ -34,10 +34,13 @@ public async Task InitializeAsync()
 
     public async ValueTask DisposeAsync()
     {
-        // prevent multiple Disposes because the Service Scope will also dispose this
-        if (_isDisposed) return;
+        lock (Instance)
```
FrankBakkerNl, line 37: Do you suspect there is a race condition here? Locking on a public object is generally not advised. Maybe this can be done with an interlocked operation on the bool is needed

src/Client/NetDaemon.HassClient/Common/HomeAssistant/Extensions/HomeAssistantConnectionHelpersExtensions.cs
```diff
@@ -6,6 +6,7 @@ public static class HomeAssistantConnectionHelpersExtensions
         this IHomeAssistantConnection connection,
         string name, CancellationToken cancelToken)
     {
+        cancelToken.ThrowIfCancellationRequested();
```
FrankBakkerNl, line 9: I think if there is no async stuff done before we get here, chances that cancellation was requested here are next to zero. It makes most sense to do this in between stuff that can take a long time, like in a loop or maybe when doing two async things after each other.

src/HassModel/NetDeamon.HassModel/Internal/AppScopedHaContextProvider.cs
```diff
@@ -100,6 +100,8 @@ public async ValueTask DisposeAsync()
     {
         if (!_tokenSource.IsCancellationRequested)
             await _tokenSource.CancelAsync();
+        //  Wait for all background tasks to complete before disposing the CancellationTokenSource
+        await _backgroundTaskTracker.DisposeAsync();
```
FrankBakkerNl, line 104: This should get disposed by the DI container. This class does not OWN it, so should not need to dispose it really.

src/AppModel/NetDaemon.AppModel/Internal/Context/ApplicationContext.cs
helto4real: Yea you are right, the boolean cannot be used since it is a value type. Yea I had errors that managed to get into this code.  I suspect there was a thread safe issue, I will just create a variable for this operation.

src/Client/NetDaemon.HassClient/Common/HomeAssistant/Extensions/HomeAssistantConnectionHelpersExtensions.cs
helto4real: In my case the code manage to call the CallService while it is cancelled. Just want to make sure it never could happen.

src/HassModel/NetDeamon.HassModel/Internal/AppScopedHaContextProvider.cs
helto4real: The dispose needs to be happening before I dispose the CancellationTokenSource or it can a problem if the corresponding token is used after. The dispose function in this case wait for every task to finnish and there could not be any problem disposing the CancellationTokenSource

helto4real: I now refactored the whole thing with a Flush method as discussed

src/Client/NetDaemon.HassClient/Common/HomeAssistant/Extensions/HomeAssistantConnectionHelpersExtensions.cs
helto4real: Removed all of them, we should be golden now given the other changes

src/AppModel/NetDaemon.AppModel/Internal/Context/ApplicationContext.cs
helto4real: I replaces the logic with interlocked mechanism. It could not be done on the boolean so had to change it to an int instead

src/Extensions/NetDaemon.Extensions.Scheduling/DisposableTimer.cs
```diff
@@ -19,6 +21,11 @@ public DisposableTimer(CancellationToken token)
         _combinedToken = CancellationTokenSource.CreateLinkedTokenSource(_internalToken.Token, token);
     }
 
+    /// <summary>
+    ///    Empty disposable timer
+    /// </summary>
+    public static IDisposable Empty => Disposable.Empty;
```
FrankBakkerNl, line 27: Why not just use this in all the places you have DisposableTimer.Empty now?

src/Extensions/NetDaemon.Extensions.Scheduling/NetDaemonScheduler.cs
```diff
@@ -40,6 +41,8 @@ public NetDaemonScheduler(ILogger<NetDaemonScheduler>? logger = null, IScheduler
     /// <inheritdoc/>
     public IDisposable RunEvery(TimeSpan timespan, Action action)
     {
+        if (_disposed) return DisposableTimer.Empty;
```
FrankBakkerNl, line 44: Because we reaaly wrap the Disposable Scheduler, which itself now handles all these auto-stopping, it seems redundant to also add these checks here.

src/HassModel/NetDaemon.HassModel.Tests/Internal/BackgroundTaskTrackerTests.cs
```diff
@@ -37,6 +37,57 @@ Task CallMe()
...
+
+        _backgroundTaskTracker.TrackBackgroundTask(CallMe(timedCancellationTokenSource.Token));
+        _backgroundTaskTracker.TrackBackgroundTask(CallMe(timedCancellationTokenSource.Token));
+        _backgroundTaskTracker.TrackBackgroundTask(CallMe(timedCancellationTokenSource.Token));
+
+        _backgroundTaskTracker.BackgroundTasks.Count.Should().Be(3);
+
+        var flushTask = _backgroundTaskTracker.Flush();
+
+        timedCancellationTokenSource.CancelAfter(500);
```
FrankBakkerNl, line 58: I think you can use a TaskCompletionSource iso a  CancellationTokenSource here so you will not need a 500ms delay in the test.

src/HassModel/NetDeamon.HassModel/Internal/BackgroundTaskTracker.cs
```diff
@@ -41,12 +46,21 @@ async Task Wrap()
         _ = Wrap();
     }
 
-    public async ValueTask DisposeAsync()
+    public async ValueTask Flush()
```
FrankBakkerNl, line 49: FlushAsync?

src/HassModel/NetDeamon.HassModel/Internal/BackgroundTaskTracker.cs
```diff
@@ -41,12 +46,21 @@ async Task Wrap()
...
     {
-        // Wait for the tasks to complete max 5 seconds
-        if (!BackgroundTasks.IsEmpty)
+        var timeoutTask = Task.Delay(TimeSpan.FromSeconds(5));
+
+        // Using a while look here incase new tasks are added while we are waiting
```
FrankBakkerNl, line 53: Did not read the comment and was really wondering why you needed the while here.. Makes sense now

src/HassModel/NetDaemon.HassModel.Tests/Internal/BackgroundTaskTrackerTests.cs
```diff
@@ -37,6 +37,57 @@ Task CallMe()
...
+    {
+        var isCalled = false;
+
+        async Task CallMe(CancellationToken cancellationToken)
+        {
+            await Task.Delay(25000, cancellationToken);
+            isCalled = true;
+        }
+
+        using var timedCancellationTokenSource = new CancellationTokenSource(25000);
```
FrankBakkerNl, line 76: Same here I think Using a TaskCompletionSource will make it independenat of timing, which also makes debugging easier

src/Extensions/NetDaemon.Extensions.Scheduling/DisposableScheduler.cs
```diff
@@ -6,26 +6,34 @@ namespace NetDaemon.Extensions.Scheduler;
...
-public sealed class DisposableScheduler : IScheduler, IDisposable
+/// <remarks>
+/// Wraps a scheduler to make it cancel all pending subscriptions on dispose.
+/// </remarks>
+public sealed class DisposableScheduler(IScheduler scheduler) : IScheduler, IDisposable
 {
-    private readonly IScheduler _innerScheduler;
+    private readonly IScheduler _innerScheduler = scheduler;
     private readonly CancellationTokenSource _cancellationTokenSource = new();
     private bool _isDisposed;
```
FrankBakkerNl, line 16: I think _disposed needs to be `volatile` here to make sure it always uses the most recent version

src/Extensions/NetDaemon.Extensions.Scheduling/DisposableTimer.cs
helto4real: It is not available in the disposable timer

src/Extensions/NetDaemon.Extensions.Scheduling/NetDaemonScheduler.cs
helto4real: Two cases, one if it is called after dispose and the other is when it is scheduled already and event happen after dispose

src/HassModel/NetDeamon.HassModel/Internal/BackgroundTaskTracker.cs
helto4real: Right!

src/HassModel/NetDaemon.HassModel.Tests/Internal/BackgroundTaskTrackerTests.cs
helto4real: Good point

src/HassModel/NetDaemon.HassModel.Tests/Internal/BackgroundTaskTrackerTests.cs
helto4real: Not sure what you mean.

src/HassModel/NetDeamon.HassModel/Internal/BackgroundTaskTracker.cs
helto4real: We resorted back to using Dispose

src/HassModel/NetDaemon.HassModel.Tests/Internal/BackgroundTaskTrackerTests.cs
helto4real: I got it now and refactored the tests

src/Extensions/NetDaemon.Extensions.Scheduling/NetDaemonScheduler.cs
helto4real: It seemed like the tests still is valid

src/Extensions/NetDaemon.Extensions.Scheduling/DisposableTimer.cs
helto4real: removed the disposable timer now

src/Extensions/NetDaemon.Extensions.Scheduling/DisposableScheduler.cs
```diff
@@ -6,26 +6,34 @@ namespace NetDaemon.Extensions.Scheduler;
...
-    }
+    private volatile bool _isDisposed;
 
     /// <inheritdoc />
     public IDisposable Schedule<TState>(TState state, Func<IScheduler, TState, IDisposable> action)
     {
         if (_isDisposed) return Disposable.Empty;
 
-        return _innerScheduler.Schedule(state, action);
+        CancellationTokenRegistration? cancellationTokenRegistration = null;
```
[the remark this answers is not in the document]
helto4real: We need it now though to be able to check the disposed flag...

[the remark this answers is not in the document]
FrankBakkerNl: Maybe this was the actual problem...

src/HassModel/NetDeamon.HassModel/Internal/QueuedObservable.cs
```diff
@@ -58,20 +58,20 @@ await foreach(var @event in  _queue.Reader.ReadAllAsync(_tokenSource.Token).Conf
...
-            
+
         // mark the channel complete, this will break the processing loop when the last event in the channel is processed
-        _queue.Writer.Complete();
-        
+        _queue.Writer.TryComplete();
+
         // now await for the processing loop to be ready so we know all pending events are processed
-        if (_eventHandlingTask != null)
+        if (_eventHandlingTask != null && !_eventHandlingTask.IsCompleted)
```
FrankBakkerNl, line 71: Seems redundant to check. await will just return directly if it was already completed

FrankBakkerNl (approved):

src/HassModel/NetDeamon.HassModel/Internal/QueuedObservable.cs
helto4real: You are right, dont even remember adding that...

PR #1135, opened 2024-07-08
Add extension function on observable to ignore OnComplete events

helto4real:
## Type of change

In certain scenarios we might want to make sure the action is not called if the underlying Observable completes. For example advanced usages of Throttle you may want to make sure it never calls the action when an application is restarted. For most cases the extension method `WhenStateIsFor` should be used that are safe for this reason but if the user still wants to use Throttle for some reason and do not want it call the action prematurely before the timeout is due, this new method can be used before the Throttle.

Example:
```csharp
ha.StateChanges()
  .Where(...)
  .IgnoreOnComplete()
  .Throttle(...)
  .Subscribe(...);
```
- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/src/HassModel/NetDeamon.HassModel/StateObservableExtensions.cs
+++ b/src/HassModel/NetDeamon.HassModel/StateObservableExtensions.cs
@@ -8,7 +8,7 @@ namespace NetDaemon.HassModel;
 public static class StateObservableExtensions
 {
     /// <summary>
-    /// Ignores the completion of the observable
+    /// Ignores the OnCompletion event of the observable
     /// </summary>
     /// <remarks>
     /// In the case you do not want the action to be called when the observable is completed
--- a/src/HassModel/NetDeamon.HassModel/StateObservableExtensions.cs
+++ b/src/HassModel/NetDeamon.HassModel/StateObservableExtensions.cs
@@ -11,28 +11,13 @@ public static class StateObservableExtensions
     /// Ignores the OnCompletion event of the observable
     /// </summary>
     /// <remarks>
-    /// In the case you do not want the action to be called when the observable is completed
-    /// use this method before the use of other methods in the Observalble.
-    /// For example if you use Throttle and you want the action call not to be
-    /// run when you cancel the observable.
-    /// For use-cases where you want to wait for a specific state for a specific time
-    /// use <see cref="WhenStateIsFor"/> instead.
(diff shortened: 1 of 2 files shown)


src/HassModel/NetDeamon.HassModel/StateObservableExtensions.cs
```diff
@@ -7,6 +7,32 @@ namespace NetDaemon.HassModel;
...
 public static class StateObservableExtensions
 {
+    /// <summary>
+    /// Ignores the OnCompletion event of the observable
+    /// </summary>
+    /// <remarks>
+    /// In the case you do not want the action to be called when the observable is completed
+    /// use this method before the use of other methods in the Observalble.
+    /// For example if you use Throttle and you want the action call not to be
+    /// run when you cancel the observable.
```
FrankBakkerNl, line 17: I think this is a bit confusing, the user will not 'cancel the observable'. Maybe better to explain the specific scenario here

`This can be used in combination with Throttle() to avoid events to be emitted when the observable is completed (eg because the app is stopped). Use IgnoreOnComplete(). before Throttle() to make sure Throttle will not send events when the app is stopping.

src/HassModel/NetDeamon.HassModel/StateObservableExtensions.cs
```diff
@@ -7,6 +7,32 @@ namespace NetDaemon.HassModel;
...
+    /// For use-cases where you want to wait for a specific state for a specific time
+    /// use <see cref="WhenStateIsFor"/> instead.
+    /// </remarks>
+    public static IObservable<T> IgnoreOnComplete<T>(this IObservable<T> source)
+        => Observable.Create<T>(observer => source.Subscribe(observer.OnNext, observer.OnError));
+
+    // public static IObservable<StateChange> IgnoreOnComplete(
```
FrankBakkerNl, line 24: Code left in comment

src/HassModel/NetDeamon.HassModel/StateObservableExtensions.cs
helto4real: Much better, thanks!

src/HassModel/NetDeamon.HassModel/StateObservableExtensions.cs
helto4real: ops :)

PR #1149, opened 2024-08-16
Improved types for generated service arguments

FrankBakkerNl:
Improved the parameter types for some generated service arguments. specifically:
date => DateOnly
time => TimeOnly
datetime => DateTime
rgb_color => IReadOnlyCollection<int>,
number without 'step' is now double instead of long

This will allow for eg:
Light.Attic.TurnOn(transition: 0.5, rgbColor: [255,128,10]);

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

- [ ] The code compiles without warnings (code quality check)


src/HassModel/NetDaemon.HassModel.Tests/CodeGenerator/ServiceMetaDataParserTest.cs
```diff
@@ -46,7 +46,7 @@ public void TestSomeBasicServicesCanBeParsed()
     [Fact]
     public void TestServicesWithAdvancedFieldsCanBeParsed()
     {
-        var sample = File.ReadAllText(@"CodeGenerator/ServiceMetaDataSamples/Lights.json");
+        var sample = File.ReadAllText(@"CodeGenerator/ServiceMetaDataSamples/light.json");
```
Barsonax, line 49: Would a raw string literal be a better way to store the json string? Then you don't have to do any file IO in the tests.

FrankBakkerNl: For these very long ones it seems to me it clutters the test to put them inline. That is arbitrary of course

Barsonax: You could put them in a separate class as constants (especially when these are used in multiple tests)

helto4real (approved):
Looks good to me, I think this breaking change is manageable and adds enough value

PR #1157, opened 2024-08-22
Properly deserialize context when deserializing state in the entity state cache

### LINKED ISSUE #1156: All values in Context are null or empty in StateChange objects
## The problem
It appears the lazy deserialization of StateChanges in 24.33.0 broke the deserialization of the Context.  In both the New and Old properties the Context has an empty string for the Id and null for ParentId and UserId.
image

## Environment
Running NetDaemon in Docker using the code in this repo [nuttytree-netdaemon](https://github.com/nuttytree/nuttytree-netdaemon) connected to Home Assistant version 2024.8.2 running in Docker.

- NetDaemon release with the issue: 24.33.0
- Last working NetDaemon release (if known): 24.28.1
- Operating environment (Home assistant Add-on/Docker/Dev setup): Docker

Looks like the issue is the result of this pull request #1146 
```c#

```

## Traceback/Error logs
```txt

```

nuttytree:
## Proposed change
Add JsonPropertyName attributes to properties in the Context object so that they properly deserialize when EntityStateCache deserializes the State object.

## Type of change
- [ ] Dependency upgrade

## Additional information
- This PR fixes or closes issue: fixes #1156 

## Checklist


FrankBakkerNl (changes requested):
I think the fix is good, was not aware of this issue. What was the actual behavior you found?

It must have been introduced with the recent lazy deserialisation of state changes.

I think the tests you updated do not really validate this fix. I even expect them to also pass without the fix because they do not depend not depend. A better test would be to deserialize actually JSON data with the names with _ in them.

Barsonax:
> It must have been introduced with the recent lazy deserialisation of state changes.

If this worked before its probably because JsonSerializerOptions used with with JsonNamingPolicy.CamelCase. If that is the case then the proper fix is to pass in those options again.

src/HassModel/NetDaemon.HassModel.Tests/Internal/EntityStateCacheTest.cs
```diff
@@ -71,6 +75,8 @@ public async Task StateChangeEventIsFirstStoredInCacheThanForwarded()
         // Assert
         eventObserverMock.Verify(m => m.OnNext(It.IsAny<HassEvent>()));
         cache.GetState(entityId)!.State.Should().Be("newState");
+        cache.GetState(entityId)!.Context!.Id.Should().NotBeNullOrEmpty();
```
helto4real, line 78: Why not just test the actual value?

helto4real:
Weird that this have never worked, seeing your fix it is obvious.

src/HassModel/NetDaemon.HassModel.Tests/Internal/EntityStateCacheTest.cs
nuttytree: Updated to test the actual value is the expected value.

nuttytree:
I have updated the tests with the requested changes.

FrankBakkerNl (approved):

src/HassModel/NetDaemon.HassModel.Tests/Internal/EntityStateCacheTest.cs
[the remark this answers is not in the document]
nuttytree: I meant to change all of those but missed that one, I updated that assert also.

helto4real:
@nuttytree thank you for your contribution. Please make sure you sign the commits in the future. It is for everyones benefit and security.

PR #1251, opened 2025-02-06
Strongly-type EntityState.Attributes

333fred:
For those working with base entities (such as the IHaRegistry API), the virtual property/covariant return override approach isn't helpful. The base property will always return a `Dictionary<string, object>`, which can be useful to iterate through. Instead, we should have strongly-typed derived class _hide_ the base property, so users who are just working with pure entities get a useful key/value pair, and derived users get strongly-typed info.


333fred:
Important note: this is a _breaking change_, so we should probably wait for a major version rev to merge?

helto4real (approved):

helto4real:
About the breaking change. It will probably not affect most people. If they subclassed it or smth it is a breaking change. There will be no supported ways to have two different versions of the apps that are compiled and the runtime anyhow. It is OK.

PR #1266, opened 2025-02-19
Use MSBuild to load projects, instead of manual parsing

333fred:
For users (like me) who use a property to specify all their netdaemon versions, instead of individual versions, the existing checks are too simple. I've updated it to instead use msbuild to actually load the project file and determine the real version that nuget will see.

### DIFF
--- a/src/HassModel/NetDaemon.HassModel.CodeGenerator/Program.cs
+++ b/src/HassModel/NetDaemon.HassModel.CodeGenerator/Program.cs
@@ -1,4 +1,5 @@
 ﻿using System.Globalization;
+using Microsoft.Build.Locator;
 using Microsoft.Extensions.Configuration;
 using NetDaemon.Client.Settings;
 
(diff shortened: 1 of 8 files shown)


src/HassModel/NetDaemon.HassModel.Tests/CodeGenerator/TestFiles/SubstitutedVersions/SubstitutedVersions.csproj
333fred, line 1: This is what I do, so I only have to update one version number and can be certain everything is in sync.

src/HassModel/NetDaemon.HassModel.CodeGenerator/Program.cs
```diff
@@ -17,6 +18,7 @@
 //This is used as a command line switch rather than a configuration key. There is no key value following it to interpret.
 generationSettings.GenerateOneFilePerEntity = args.Any(arg => arg.ToLower(CultureInfo.InvariantCulture) == "-fpe");
 
+MSBuildLocator.RegisterDefaults();
```
333fred, line 21: Note that this does mean that anyone who runs `nd-codegen` _must_ have MSBuild available. Since I only expect this to be run from a development machine, I don't particularly see this as concerning.

FrankBakkerNl (approved):

Issue #1306, opened 2025-05-06
JSONException after upgrade to 25.18.0


bramnauta:
## The problem
After upgrading to  25.18.0, reloading NetDaemon fails with an exception (see below). After downgrading the NuGet packages to 25.14.1, everything started working again. (Addon is still at 25.18.0)
I don't have the time to debug further right now, but wanted to report this in case it's something that'll break upgrading for more people. I will look into this further tomorrow. Also if there is anything else you need from me, feel free to let me know so I can provide that asap tomorrow.

## Environment

Addon: 25.18.0
NuGet packages: 25.18.0

- NetDaemon release with the issue: 25.18.0
- Last working NetDaemon release (if known): 25.14.1
- Operating environment (Home assistant Add-on/Docker/Dev setup): Home Assistant OS with NetDaemon addon
- Link to integration documentation on our website:

## Link to or paste code that causes the issue

```txt
(NetDaemon startup)
```

## Traceback/Error logs

```txt
[22:19:49 INF] NetDaemon.Runtime.Internal.NetDaemonRuntime: Successfully connected to Home Assistant
[22:19:50 INF] NetDaemon.HassModel.Internal.RegistryCache: Initializing RegistryCache
[22:19:50 FTL] NetDaemon.Runtime.Internal.NetDaemonRuntime: Error (re-)initializing after connect to Home Assistant
System.Text.Json.JsonException: The JSON value could not be converted to System.String. Path: $[88].identifiers[0][1] | LineNumber: 0 | BytePositionInLine: 58780.
 ---> System.InvalidOperationException: Cannot get the value of a token type 'Number' as a string.
   at System.Text.Json.ThrowHelper.ThrowInvalidOperationException_ExpectedString(JsonTokenType tokenType)
   at System.Text.Json.Utf8JsonReader.GetString()
   at System.Text.Json.Serialization.JsonCollectionConverter`2.OnTryRead(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options, ReadStack& state, TCollection& value)
   at System.Text.Json.Serialization.JsonConverter`1.TryRead(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options, ReadStack& state, T& value, Boolean& isPopulatedValue)
   at System.Text.Json.Serialization.JsonCollectionConverter`2.OnTryRead(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options, ReadStack& state, TCollection& value)
[...]

PR #1309, opened 2025-05-10
Fix crash in 2025.5

333fred:
Fixes https://github.com/net-daemon/netdaemon/issues/1306 by ensuring that all identifiers are converted to strings.

### DIFF
--- a/src/HassModel/NetDaemon.HassModel.Tests/Internal/HassObjectMapperTest.cs
+++ b/src/HassModel/NetDaemon.HassModel.Tests/Internal/HassObjectMapperTest.cs
@@ -43,7 +43,9 @@ public class HassObjectMapperTest
                 [
                     "SKIP"
                 ],
-                [1,2]
+                [1,2],
+                [1.1, 2],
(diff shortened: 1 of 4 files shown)


src/HassModel/NetDaemon.HassModel.Tests/Internal/HassObjectMapperTest.cs
```diff
@@ -14,7 +14,10 @@ public class HassObjectMapperTest
     private readonly JsonSerializerOptions _defaultSerializerOptions = new()
     {
         WriteIndented = false,
-        DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull
+        DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
+
+        // HACK: this was just a way to verify it works when adding the converter here
+        Converters = { new EnsureStringConverter() }
```
FrankBakkerNl, line 20: This was not intended to be left here

FrankBakkerNl (approved):

PR #1389, opened 2026-06-18
[codex] Improve high-event hot paths

helto4real:
## Breaking change
None. This keeps public APIs unchanged and preserves raw `OnHassMessage` publication behavior.

## Proposed change
This PR implements the measured high-event performance improvements from the performance review plan.

The command/result hot path now uses an ID-indexed pending-result table instead of creating one filtered Rx subscription per command. `HomeAssistantConnection` registers a `TaskCompletionSource<HassMessage>` for each command id, completes matching `result` messages directly in the receive loop, and still publishes every raw `HassMessage` to `OnHassMessage` for ping, trigger, and diagnostic consumers.

Review follow-up fixed the pending-result lifecycle: pending entries are removed when a result arrives, the connection is disposed, sending fails, the caller cancels while waiting, or the result wait times out. Tests cover timeout and cancellation cleanup.

The websocket receive path reads a complete websocket message into a byte buffer, detects whether the payload is a JSON object or coalesced array, and deserializes directly to `T` or `T[]`. This removes the prior `JsonElement` intermediate and second typed deserialization on the ingest path while preserving chunked message and remote-close behavior.

The benchmark harness keeps the owning `JsonDocument` alive for the state-change benchmark lifetime, fixing the previous invalid `JsonElement` baseline failure.

Review follow-up also corrected `ResultDispatchBenchmarks` so the implemented path uses `ConcurrentDictionary` and awaits all completion tasks, matching the Rx baseline's completed-work shape. The corrected local in-process result-dispatch run showed the expected fixed overhead for one pending command and large wins as pending command count grows:

| Pending commands | Baseline Rx dispatch | Concurrent dictionary dispatch | Result |
| ---: | ---: | ---: | --- |
| 1 | 250.2 ns, 880 B | 485.2 ns, 1,800 B | slower for a single waiter |
| 10 | 2,175.3 ns, 6,424 B | 929.4 ns, 3,312 B | 57% faster, 48% lower allocation |
| 100 | 30,939.1 ns, 133,144 B | 7,566.0 ns, 24,880 B | 76% faster, 81% lower allocation |
| 1000 | 1,819.0 us, 8,528,351 B | 80.2 us, 262,848 B | 96% faster, 97% lower allocation |

Detailed local report: `docs/performance/high-event-review.md`.

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

No user-facing functionality or configuration was added or changed.


src/Client/NetDaemon.HassClient/Internal/HomeAssistantConnection.cs
```diff
@@ -189,17 +193,31 @@ private async Task<Task<HassMessage>> SendCommandAsyncInternal<T>(T command, Can
...
-            var resultEvent = _hassMessageSubject
-                .Where(n => n.Type == "result" && n.Id == command.Id)
-                .FirstAsync().ToTask(CancellationToken.None);
-            // We dont want to pass the incoming CancellationToken here because it will throw a TaskCanceledException
-            // when calling services from an Apps Dispose(Async) and hide possible actual exceptions
+            // Complete result waiters directly from the receive loop instead of creating
+            // one filtered Rx subscription per command.
+            var resultEvent = new TaskCompletionSource<HassMessage>(
+                TaskCreationOptions.RunContinuationsAsynchronously);
+            if (!_pendingResults.TryAdd(command.Id, resultEvent))
```
helto4real, line 208: This entry is removed when a result arrives or during disposal, but not when the caller's `WaitAsync` times out or is cancelled. If Home Assistant never returns the result while the connection stays alive, every failed command retains a TCS and dictionary entry indefinitely. Please remove it on timeout/cancellation and add tests for both paths.

src/Client/NetDaemon.HassClient/Internal/HomeAssistantConnection.cs
```diff
@@ -189,17 +193,31 @@ private async Task<Task<HassMessage>> SendCommandAsyncInternal<T>(T command, Can
...
+            catch (OperationCanceledException)
+            {
+                if (_pendingResults.TryRemove(command.Id, out var pendingResult))
+                    pendingResult.TrySetCanceled(cancelToken);
+                throw;
+            }
+            catch (Exception e)
+            {
+                if (_pendingResults.TryRemove(command.Id, out var pendingResult))
+                    pendingResult.TrySetException(e);
```
helto4real: This TCS task is never returned when `SendMessageAsync` throws—the enclosing method rethrows instead—so faulting it creates an unobserved faulted task. Removing the entry from `_pendingResults` and rethrowing is sufficient unless the TCS task is explicitly observed.

benchmarks/NetDaemon.PerformanceBenchmarks/Benchmarks/ResultDispatchBenchmarks.cs
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        await Task.WhenAll(tasks).ConfigureAwait(false);
+    }
+
+    [Benchmark]
+    public void DictionaryResultDispatch()
+    {
+        var pending = Enumerable.Range(1, PendingCommands)
+            .ToDictionary(static id => id, static _ => new TaskCompletionSource<HassMessage>(TaskCreationOptions.RunContinuationsAsynchronously));
```
helto4real: This benchmark uses a regular `Dictionary`, while production uses `ConcurrentDictionary`, and unlike the Rx baseline it never awaits the TCS tasks. It therefore measures less work and a different data structure. Please benchmark equivalent end-to-end production paths and await completion symmetrically before using the reported 79–98% improvement.

helto4real (commented):
Code review summary: the optimization direction looks sound and CI is green, but the pending-result lifecycle should be fixed before merging, and the benchmark should compare equivalent production paths before the precise performance claims are retained.

FrankBakkerNl (approved):

PR #1445, opened 2026-09-05
Fix inverted app enable/disable state write

DevJasperNL:
## Proposed change

`AppStateManager.SaveStateAsync` only calls `IAppStateRepository.UpdateAsync` when the requested `ApplicationState` differs from the current `input_boolean` state, but it passed the *current* `isEnabled` value instead of the desired one. As a result every real write did the opposite of what was requested:

- `SaveStateAsync(id, Enabled)` while the helper is `off` sent `input_boolean.turn_off`
- `SaveStateAsync(id, Disabled)` while the helper is `on` sent `input_boolean.turn_on`

This PR passes `state == ApplicationState.Enabled` to `UpdateAsync` so the helper is set to the desired state.

The existing `AppStateManagerTests` only verified that *some* `call_service` command was sent, which is why this slipped through. The two affected tests now assert the exact domain and service (`input_boolean.turn_off` / `input_boolean.turn_on`) and that it is sent exactly once. Both tests fail against the previous code and pass with the fix.

## Type of change

- [ ] Dependency upgrade

## Additional information

Found during a code review of the runtime; no issue was filed for it. The bug only affects users who opt in via `AddNetDaemonStateManager()`.

## Checklist

### DIFF
--- a/src/Runtime/NetDaemon.Runtime.Tests/Internal/AppStateManagerTests.cs
+++ b/src/Runtime/NetDaemon.Runtime.Tests/Internal/AppStateManagerTests.cs
@@ -82,10 +82,10 @@ public class AppStateManagerTests
         // ASSERT
         haConnectionMock.Verify(n =>
             n.GetApiCallAsync<HassState>("states/input_boolean.netdaemon_helloapp", It.IsAny<CancellationToken>()));
-        // It exists so it should turn it on
+        // It is on so it should turn it off
         haConnectionMock.Verify(n =>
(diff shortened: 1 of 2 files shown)


helto4real (approved):
The fix correctly persists the requested enabled state, and the strengthened assertions cover both turn_on and turn_off. All 50 runtime tests passed on the reviewed pre-rebase commit; the diff is unchanged after rebasing onto current main. Approved, with merge pending the required CI checks on the rebased commit.

PR #1447, opened 2026-09-05
Skip service domains without services in the code generator

DevJasperNL:
## Proposed change

`nd-codegen` could emit a `HomeAssistantGenerated.cs` that does not compile.

`Generator.GenerateTypes` handed every service domain to `HelpersGenerator` and `ServicesGenerator`. `ServicesGenerator` only emitted a `{Domain}Services` class for domains that still had at least one service, but it built the `IServices` interface and the `Services` root class from *all* domains, and `HelpersGenerator` registered `AddTransient<{Domain}Services>()` for all domains as well.

`ServiceMetaDataParser` swallows per-service deserialization errors by design. When every service of a domain fails to parse, that domain reaches the generators with an empty service list. The generated file then contained an `IServices` property, a `Services` property and a DI registration for a `{Domain}Services` type that was never declared, so the user's project no longer built.

This PR:

- Filters out domains without services once, in `Generator.GenerateTypes`, before the helpers, services and extension-method generators see them. All generated types now agree on the same set of domains.
- Removes the local `Where(sd => sd.Services.Any())` in `ServicesGenerator`, which was the source of the inconsistency. Given the same input, the root interface, the root class and the domain classes are now always consistent.
- Adds `TestDomainWithoutAnyServices_ServicesTypeNotReferenced` to `ServicesGeneratorTest`. It generates code for a domain with zero services next to a normal `light` domain, asserts the `{Domain}Services` name does not appear anywhere in the output, and compiles the generated code together with app code. It fails without the fix with three dangling references to the undeclared type.

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/src/HassModel/NetDaemon.HassModel.CodeGenerator/CodeGeneration/Generator.cs
+++ b/src/HassModel/NetDaemon.HassModel.CodeGenerator/CodeGeneration/Generator.cs
@@ -13,7 +13,9 @@ internal static class Generator
         IReadOnlyCollection<EntityDomainMetadata> entityDomains,
         IReadOnlyCollection<HassServiceDomain> services)
     {
-        var orderedServiceDomains = services.OrderBy(x => x.Domain).ToArray();
+        // Domains whose services all failed to parse have no services class, so they must not be
+        // referenced from IServices, Services or the DI registrations either
+        var orderedServiceDomains = services.Where(x => x.Services.Any()).OrderBy(x => x.Domain).ToArray();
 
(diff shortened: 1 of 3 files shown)


FrankBakkerNl (approved):

helto4real (approved):
The shared filtering keeps the service interfaces, root types, domain classes and DI registrations consistent. All 153 HassModel tests passed locally, and the new regression test failed as expected when the production fix was removed. The reviewed HassModel code and tests are unchanged after rebasing onto main. Approved, with merge pending required CI on the rebased commit.

PR #1448, opened 2026-09-05
Return null from RegistryCache.GetLabelById for unknown labels

DevJasperNL:
## Proposed change

`RegistryCache.GetLabelById` used the dictionary indexer, so looking up a label id that is not in the label cache threw a `KeyNotFoundException`. Every sibling lookup (`GetHassEntityById`, `GetDeviceById`, `GetAreaById`, `GetFloorById`) already returns `null` for unknown ids, and `HassObjectMapper` filters those nulls out with `OfType<Label>()` when it maps the `Labels` collections of entities, devices and areas.

Each registry (entities, devices, areas, floors, labels) is reloaded independently and each has its own 5-minute throttle on its `*_registry_updated` event. This means the entity or area registry can reference a label the label registry has not loaded yet, for example right after a label is created and assigned in the Home Assistant UI. Any app that then touched `IHaContext.Registry`, `entity.Registration`, or navigated to an `Area`/`Device` threw instead of seeing the entity without that label.

This PR:

- Changes `GetLabelById` to use `GetValueOrDefault`, in line with the other lookups. An unknown label is now simply omitted from the mapped `Labels` lists, and `IHaRegistry.GetLabel` returns `null` for it.
- Adds `TestUnknownLabelIsIgnored` to `RegistryNavigationTest`. An entity and an area reference a label id that is not in the label registry; the test asserts `GetLabel` returns `null`, the entity registration keeps only the known label, and the area has no labels. Without the fix it fails with `KeyNotFoundException: The given key 'unknown_label' was not present in the dictionary.`

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/src/HassModel/NetDaemon.HassModel.Tests/Registry/RegistryNavigationTest.cs
+++ b/src/HassModel/NetDaemon.HassModel.Tests/Registry/RegistryNavigationTest.cs
@@ -152,6 +152,40 @@ public class RegistryNavigationTest
     }
 
     [Fact]
+    public async Task TestUnknownLabelIsIgnored()
+    {
+        // Setup: the label registry can lag behind the entity/area registries because each one is
+        // reloaded independently, so references to labels that are not (yet) known must not throw
+        SetupCommandResult("config/entity_registry/list",
+            [
+                new HassEntity { EntityId = "light.babyroom_nightlight", AreaId = "baby_room", Labels = ["stay_on", "unknown_label"] },
+            ]);
+        SetupCommandResult("config/device_registry/list", Array.Empty<HassDevice>());
+        SetupCommandResult("config/area_registry/list",
+            [
+                new HassArea { Name = "Baby room", Id = "baby_room", Labels = ["unknown_label"] },
+            ]);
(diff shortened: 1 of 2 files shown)


helto4real:
if you are using an agent for these fixes please make it use correct label

FrankBakkerNl (approved):

helto4real (changes requested):
The GetValueOrDefault fix is correct and the unknown-label regression test works: all 153 HassModel tests passed locally, and reverting the production change made the new test fail with KeyNotFoundException. Before approval, please add a focused assertion/test that RegistryCache.GetLabelById(null) returns null (or exercise that path through the concrete HaRegistry.GetLabel method). The modified line currently has only partial branch coverage: local coverage recorded 19 executions but zero hits for the null-id branch, consistent with the failing codecov/patch check. Keep the existing known-label and unknown-label assertions, rerun the HassModel suite with coverage, and ensure codecov/patch is green. No production-code change or coverage-threshold reduction is needed; this request is for the missing null-input test.

helto4real (approved):

